PR #10, opened 2017-04-24
Added filter push up test

erkkel:
The pushup filter test now matches directly against the physical plan


ankurdave:
Closing this for now since we have removed oblivious mode from the public release.

PR #12, opened 2017-05-04
Unified serialization and evaluation using Flatbuffers

### LINKED ISSUE #13: Remove dependency on SGX at the client
Boundary processing for join and aggregation currently occur at the client using a call into the SGX enclave. These should instead use a single worker partition, allowing the client to run on a non-SGX-enabled machine.

ankurdave:
This PR addresses two problems in Opaque:
- Brittle serialization logic: Rows are currently serialized using error-prone custom byte manipulation logic. This has led to a number of problems when handling edge cases like empty partitions.
- Query inflexibility: Operators and expressions are indicated using opcodes that represent specific, hardcoded queries. This makes it very tedious to express new queries.

This PR solves these problems using Flatbuffers, a fast serialization library with explicit schemas.
- Flatbuffers serves as a unified serialization format for rows, operators, and expressions. The PR includes schemas for each of these.
- The enclave contains an interpreter that can handle arbitrary expressions serialized using this format.

Because we implement boundary processing for join and aggregation using pure Scala, this fixes #13.

It will be ready to merge when it reaches feature parity with the current version of Opaque. Checklist:
  - [x] Project
  - [x] Filter
  - [x] Single-partition sort
  - [x] Distributed sort
  - [x] Single-partition inner equi-join
  - [x] Distributed inner equi-join
  - [x] Aggregate


ankurdave:
Hooray!

PR #22, opened 2018-03-07
Add support to SGX 2.0

oshritf:
Makefile adjustment to enable Opaque to run with SGX2.0
Tested with "Big Data Benchmark" suite on Linux SGX 2.0 sdk environment


oshritf:
To be checked with SGX2.0 SDK installed

wzheng:
Thank you!

PR #30, opened 2018-03-19
Truce integration

ggershinsky:


ankurdave:
Thanks for this! I finally got a chance to try it out. I found setting up TruCE's dependencies, especially SSL SGX, to be pretty involved and more than we would want for a first-time user of Opaque.

Maybe it would be better to make TruCE an optional dependency so that users only have to install it when they actually deploy Opaque to the untrusted cloud. I can modify the PR to separate out TruCE from the build.

PR #50, opened 2018-11-06
Prevent Spark Shell From Exiting Upon Exception

### LINKED ISSUE #27: Native failed checks result in program exit
Enclave check method uses std::exit if condition is not met.
As a result, each time an enclave check method fails, for example in data manipulation such as join or filter, the entire application is exited and has to be restarted.

chester-leung:
This PR introduces logic that prevents the Spark Shell from exiting upon an exception. Opaque now handles enclave errors by catching C++ exceptions and rethrowing them with the JNI. 

Also introduces `OpaqueException`, a type of Java exception representing an exception occurring from within an Opaque enclave. 

Refactors some code to throw C++ runtime errors that are caught in `ecall`s, which then proceed to throw `OpaqueException`s. Previously, Opaque would simply exit out of the Spark Shell.

Adds a test to test this functionality.

Co-authored by @ankurdave 

Fixes #27


ankurdave:
Thanks!

PR #53, opened 2018-11-09
Upgrade to SBT 0.13.17 for Java 9+ support

ankurdave:


ankurdave:
It seems this still won't actually provide Java 9+ support, because Spark does not support Java 9+: [SPARK-24421](https://issues.apache.org/jira/browse/SPARK-24421).

PR #60, opened 2018-11-12
Support schema inference when loading encrypted data

### LINKED ISSUE #59: Support schema inference when loading encrypted data
`EncryptedSource` currently extends `SchemaRelationProvider`, meaning the user must provide a schema when loading an encrypted table. This is inconvenient when accessing Opaque through pure SQL.

We should instead serialize the schema alongside the encrypted data, make `EncryptedSource` extend `RelationProvider` to remove the need for a user-specified schema, and load the saved schema in `EncryptedSource#createRelation`.

ankurdave:
Schema inference is implemented by serializing the schema alongside the encrypted data. Schemas are saved to `${path}/schema`, while data is saved to `${path}/data`.

Fixes #59.

Co-authored-by: Chester Leung <chester.leung@berkeley.edu>

### DIFF
--- a/src/main/scala/edu/berkeley/cs/rise/opaque/sources.scala
+++ b/src/main/scala/edu/berkeley/cs/rise/opaque/sources.scala
@@ -17,25 +17,46 @@
 
 package edu.berkeley.cs.rise.opaque
 
+import java.io.File
+import java.io.FileInputStream
+import java.io.FileOutputStream
+import java.io.ObjectInputStream
+import java.io.ObjectOutputStream
+
 import org.apache.spark.rdd.RDD
+import org.apache.spark.sql.DataFrame
 import org.apache.spark.sql.SQLContext
 import org.apache.spark.sql.SaveMode
 import org.apache.spark.sql.SparkSession
-import org.apache.spark.sql.DataFrame
 import org.apache.spark.sql.sources.BaseRelation
 import org.apache.spark.sql.sources.CreatableRelationProvider
+import org.apache.spark.sql.sources.RelationProvider
 import org.apache.spark.sql.sources.SchemaRelationProvider
 import org.apache.spark.sql.types.StructType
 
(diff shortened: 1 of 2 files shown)


src/main/scala/edu/berkeley/cs/rise/opaque/sources.scala
```diff
@@ -17,25 +17,46 @@
...
+class EncryptedSource
+    extends RelationProvider
+    with SchemaRelationProvider
+    with CreatableRelationProvider {
+
+  override def createRelation(
+    sqlContext: SQLContext,
+    parameters: Map[String, String]): BaseRelation = {
+    val schemaFile = new File(parameters("path"), "schema")
+    val schema = new ObjectInputStream(new FileInputStream(schemaFile))
```
ankurdave: We should instead use the Hadoop APIs for writing the schema, since `parameters("path")` may be a Hadoop or S3 path.

src/main/scala/edu/berkeley/cs/rise/opaque/sources.scala
```diff
@@ -46,8 +67,16 @@ class EncryptedSource extends SchemaRelationProvider with CreatableRelationProvi
...
     val blocks: RDD[Block] = data.queryExecution.executedPlan.asInstanceOf[OpaqueOperatorExec]
       .executeBlocked()
-    blocks.map(block => (0, block.bytes)).saveAsSequenceFile(parameters("path"))
-    EncryptedScan(parameters("path"), data.schema)(
+
+    val dataDir = new File(parameters("path"), "data")
+    blocks.map(block => (0, block.bytes)).saveAsSequenceFile(dataDir.getPath)
+
+    val schemaFile = new File(parameters("path"), "schema")
+    val os = new ObjectOutputStream(new FileOutputStream(schemaFile))
```
ankurdave: Also here.

Issue #64, opened 2018-11-15
Remove client-side enclave dependency


ankurdave:
The client uses `ecall_encrypt`, `ecall_decrypt`, and `ecall_find_range_bounds`. The first two should be reimplemented using a Java AES library. The latter should be run on a worker instead of the driver.

After this is done, then `ecall_decrypt` can be removed (#37).

PR #65, opened 2018-11-19
Non-enclave Dependent Encrypt/Decrypt

### LINKED ISSUE #37: Remove ecall_decrypt from server side
The capability to decrypt arbitrary data on the server side was added for debugging only, but was never removed before Opaque's release. With this debugging function, an attacker can decrypt arbitrary data, breaking Opaque's security model.

No server-side code calls ecall_decrypt, but it is [used](https://github.com/ucbrise/opaque/blob/0b8e5eae108c96c485f77b7d0bdd2f056c02456c/src/main/scala/edu/berkeley/cs/rise/opaque/execution/operators.scala#L148) on the client to decrypt the results of an Opaque query. Two changes are therefore needed:

1. On the server side, ecall_decrypt should be removed from the enclave interface.
2. On the client side, calls to ecall_decrypt should be replaced with a [different crypto implementation](https://github.com/ucbrise/opaque/blob/0b8e5eae108c96c485f77b7d0bdd2f056c02456c/src/enclave/ServiceProvider/sp_crypto.h#L295) that does not depend on the presence of SGX hardware.

chester-leung:
Add `encrypt()` and `decrypt()` util functions to remove client side enclave dependency when encrypting and decrypting

Reorder outputted cipher in enclave crypto functions to match Java's ordering. Was previously `(IV, MAC, data)`, is now `(IV, data, MAC)`

Co-authored by @ankurdave

Fixes #37 
WIP on #64


ankurdave:
Thanks!

PR #78, opened 2019-03-12
Add logistic regression benchmark

ankurdave:
This PR adds the following expressions and UDAFs to Opaque:

- Exp (expression - exponential function)
- VectorAdd (expression - adds two vectors elementwise, used to implement the VectorSum UDAF)
- VectorMultiply (expression - multiplies a vector by a scalar)
- DotProduct (expression - computes the dot product of two vectors)
- UnaryMinus (expression - implemented in terms of Subtract)
- VectorSum (UDAF - sums vectors elementwise)

It then implements logistic regression in Spark SQL using these expressions and UDAFs. The implementation follows the [Spark example](https://github.com/apache/spark/blob/b8dd84b9e4690efa9773f73787e19a4d60257e5e/examples/src/main/scala/org/apache/spark/examples/SparkLR.scala).

Also, it adds the ability to compare Array[Double] using a numeric tolerance in OpaqueOperatorTests.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -12,6 +12,8 @@ sparkVersion := "2.4.0"
 
 sparkComponents ++= Seq("core", "sql", "catalyst")
 
+libraryDependencies += "org.scalanlp" %% "breeze" % "0.13.2"
+
 libraryDependencies += "org.scalatest" %% "scalatest" % "2.2.6" % "test"
 
 val flatbuffersVersion = "1.7.0"
--- a/build.sbt
+++ b/build.sbt
@@ -14,7 +14,7 @@ sparkComponents ++= Seq("core", "sql", "catalyst")
 
 libraryDependencies += "org.scalanlp" %% "breeze" % "0.13.2"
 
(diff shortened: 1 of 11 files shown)


mirsci:
Hi @ankurdave, 

great work and design with Opaque oblivious operators to abstract the SGX enclave interactions!
 
Would you consider adding more statistical/machine learning algorithms, beside logistic regression, in the future within Opaque? In other words, extending Opaque with ML models is included in its future roadmap?

Thanks in advance!

ankurdave:
Thanks!

Logistic regression was a specific request from a prospective user, and turned out to be easy to express in SQL. We don't currently have the cycles to implement a wide range of new ML algorithms, but are there any you would suggest in particular?

mirsci:
Thanks @ankurdave, for your quick response. If we can start with any of the clustering models (e.g. k-means), it will be awesome. 
Will this be possible within Opaque?

ankurdave:
Thanks for the suggestion. I just implemented k-means in #81.

mirsci:
This is awesome, @ankurdave , will try it out. Thanks!

PR #91, opened 2020-06-11
Document missing steps in remote attestation

ankurdave:
The current implementation of remote attestation verifies that the enclave is genuine, but it does not verify that (1) it is running the correct code, and (2) it is in release mode. (Launching enclaves in release mode requires a [Launch Enclave](https://github.com/intel/linux-sgx/blob/master/psw/ae/ref_le/ref_le.md) or a license from Intel.)

In addition, the enclave is currently not signed using the user's private key, but rather a sample key. This will need to change once we start validating the enclave's identity.

This PR adds TODOs to document these missing steps.


wzheng:
Thanks!

PR #96, opened 2020-09-05
Fix comparison function handle NULL values

wzheng:
This PR adds a fix to enable comparison with NULL values, which is used by all of the sort-based operators. The previous comparison function returned NULL when either of the input values is NULL, but also left the underlying value of the field to be false. While this null-intolerant behavior is correct for projection, the underlying value is incorrectly used by Opaque’s internal sorting function, which causes a wrong sort ordering when the input data contains NULL values.

### DIFF
--- a/src/enclave/Enclave/ExpressionEvaluation.h
+++ b/src/enclave/Enclave/ExpressionEvaluation.h
@@ -102,13 +102,17 @@ flatbuffers::Offset<tuix::Field> eval_binary_arithmetic_op(
  *
  * The left and right Fields are the inputs to the binary operation. They may be temporary pointers
  * invalidated by further writes to builder; this function will not read them after invalidating.
+ * 
+ * When either of the values is NULL, this function will return a NULL value, with a boolean underlying 
+ * value that is used for internal functions like sorting. 
+ * The parameter `nulls_first` enables the function to order either by NULLS FIRST or by NULLS LAST.
  */
 template<typename TuixExpr, template<typename T> class Operation>
 flatbuffers::Offset<tuix::Field> eval_binary_comparison(
   flatbuffers::FlatBufferBuilder &builder,
   const tuix::Field *left,
   const tuix::Field *right,
-  bool is_null_asc = true) {
+  bool nulls_first = true) {
 
   if (left->value_type() != right->value_type()) {
     throw std::runtime_error(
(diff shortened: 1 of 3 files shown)


src/enclave/Enclave/ExpressionEvaluation.h
```diff
@@ -107,7 +107,8 @@ template<typename TuixExpr, template<typename T> class Operation>
 flatbuffers::Offset<tuix::Field> eval_binary_comparison(
   flatbuffers::FlatBufferBuilder &builder,
   const tuix::Field *left,
-  const tuix::Field *right) {
+  const tuix::Field *right,
+  bool is_null_asc = true) {
```
ankurdave, line 111: It would be good to mention this `is_null_asc` parameter in the method description, particularly because it does something surprising: it uses the underlying value of a NULL field to return extra information.

ankurdave: Also, the parameter name is a little confusing. How about `nulls_first`? This follows the Spark SQL [NullOrdering](https://github.com/apache/spark/blob/0b3bb45b893005d73afb15e4904ac6440c87811c/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/SortOrder.scala#L47) naming.

src/enclave/Enclave/ExpressionEvaluation.h
```diff
@@ -205,6 +206,18 @@ flatbuffers::Offset<tuix::Field> eval_binary_comparison(
         + std::string(" on ")
         + std::string(tuix::EnumNameFieldUnion(left->value_type())));
     }
+  } else {
+    // This code block handles comparison when at least one value is NULL.
+    // The logic can be summarized as: (note that the != operation implements XOR for boolean values)
+    // If is_null_asc = 1
+    // (x, NULL) = (0, 1) => (1, 0), (NULL, x) => (0, 1), (NULL, NULL) => (0, 0)
```
ankurdave, line 213: I'm not following the meaning of this line. Perhaps it would be better expressed as a table? Alternatively we could remove it.

src/test/scala/edu/berkeley/cs/rise/opaque/OpaqueOperatorTests.scala
```diff
@@ -283,6 +283,17 @@ trait OpaqueOperatorTests extends FunSuite with BeforeAndAfterAll { self =>
     df.sort($"x", $"y").collect
   }
 
+  testAgainstSpark("sort with null values") { securityLevel =>
+    val data : Seq[Tuple1[Integer]] = Random.shuffle((0 until 256).map(x => {
```
ankurdave, line 287: Style: No space before the `:` ([ref](https://docs.scala-lang.org/style/types.html#annotations))

src/enclave/Enclave/ExpressionEvaluation.h
ankurdave: It would also be good to call out the surprising use of this method at the call site where it occurs ([`FlatbuffersSortOrderEvaluator::less_than()`](https://github.com/mc2-project/opaque/blob/e890e2aae4751a215ce43cd47c1168b67077353c/src/enclave/Enclave/ExpressionEvaluation.h#L1137)). Something like "We ignore the result's NULL flag and use only its underlying value for the comparison. `eval_binary_comparison()` uses this underlying value to pass the desired information."

src/enclave/Enclave/ExpressionEvaluation.h
wzheng: Changed to a table.

src/enclave/Enclave/ExpressionEvaluation.h
wzheng: Improved comments.

ankurdave (approved):
Very nice!

chester-leung (approved):

PR #99, opened 2020-09-09
Fix filter bug

wzheng:
This PR fixes a bug in filter, where it throws an error when filtering on a `NULL` value. This changes to behavior to return `false` instead.

### DIFF
--- a/src/enclave/Enclave/Filter.cpp
+++ b/src/enclave/Enclave/Filter.cpp
@@ -26,7 +26,8 @@ void filter(uint8_t *condition, size_t condition_length,
     }
 
     // If condition_result is NULL, then always return false
-    bool keep_row = static_cast<const tuix::BooleanField *>(condition_result->value())->value() & !condition_result->is_null();
+    bool keep_row = !condition_result->is_null() &&
+      static_cast<const tuix::BooleanField *>(condition_result->value())->value();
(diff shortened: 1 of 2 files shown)


src/enclave/Enclave/Filter.cpp
```diff
@@ -24,11 +24,9 @@ void filter(uint8_t *condition, size_t condition_length,
         std::string("Filter expression expected to return BooleanField, instead returned ")
         + std::string(tuix::EnumNameFieldUnion(condition_result->value_type())));
     }
-    if (condition_result->is_null()) {
-      throw std::runtime_error("Filter expression returned null");
-    }
 
-    bool keep_row = static_cast<const tuix::BooleanField *>(condition_result->value())->value();
+    // If condition_result is NULL, then always return false
+    bool keep_row = static_cast<const tuix::BooleanField *>(condition_result->value())->value() & !condition_result->is_null();
```
ankurdave, line 29: 1. Best to keep a maximum line length of 100 characters to match Spark and the rest of the C++ codebase.
2. Might as well use logical AND here, unless there's a specific reason to use bitwise AND?

ankurdave (approved):

chester-leung (approved):
LGTM. Thanks!

PR #103, opened 2020-09-17
add between test

eric-feng-2011:
Add test for BETWEEN expression. Opaque already supports BETWEEN, as it's a a shorthand for `>= AND <=`, but there was previously no test for it.

### DIFF
--- a/src/test/scala/edu/berkeley/cs/rise/opaque/OpaqueOperatorTests.scala
+++ b/src/test/scala/edu/berkeley/cs/rise/opaque/OpaqueOperatorTests.scala
@@ -411,6 +411,12 @@ trait OpaqueOperatorTests extends FunSuite with BeforeAndAfterAll { self =>
     df.filter($"word".contains(lit("1"))).collect
   }
 
+  testAgainstSpark("between") { securityLevel =>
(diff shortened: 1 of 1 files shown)


wzheng:
Can you write a quick description of what this PR is about? Otherwise looks good!

src/test/scala/edu/berkeley/cs/rise/opaque/OpaqueOperatorTests.scala
```diff
@@ -411,6 +411,12 @@ trait OpaqueOperatorTests extends FunSuite with BeforeAndAfterAll { self =>
     df.filter($"word".contains(lit("1"))).collect
   }
 
+  testAgainstSpark("between") { securityLevel =>
+    val data = for (i <- 0 until 256) yield(i.toString, i)
+    val df = makeDF(data, securityLevel, "word", "count")
+    df.filter($"count".between(3, 5)).collect
```
chester-leung, line 417: Can you increase the range of the between? Maybe something like `between(50, 150)`

eric-feng-2011: Changed

chester-leung:
Thanks!

Issue #26, opened 2018-03-08
Memory leak in FlatbuffersTemporaryRow


ankurdave:
`FlatbuffersTemporaryRow::set` [acquires ownership](https://github.com/ucbrise/opaque/blob/80346b7085043cb9acd292514b80395bce424a74/src/enclave/Enclave/Flatbuffers.h#L434) of a Rows buffer via a raw pointer but never frees the pointer, creating a memory leak when `set` is called again.

To prevent this kind of memory leak, [`FlatbuffersRowWriter::output_buffer`](https://github.com/ucbrise/opaque/blob/80346b7085043cb9acd292514b80395bce424a74/src/enclave/Enclave/Flatbuffers.h#L388) should return a smart pointer instead of a raw pointer. We will need a custom smart pointer because the managed memory was allocated outside of the enclave and must be deallocated using `ocall_free`.

PR #104, opened 2020-09-24
Upper implementation and test

eric-feng-2011:
Upper implementation and test. This commit removes the docker readme, because git ...

### DIFF
--- a/src/enclave/Enclave/ExpressionEvaluation.h
+++ b/src/enclave/Enclave/ExpressionEvaluation.h
@@ -722,6 +722,58 @@ private:
       }
     }
 
+    case tuix::ExprUnion_Upper:
+    {
+      auto n = static_cast<const tuix::Upper *>(expr->expr());
+      auto child_offset = eval_helper(row, n->child());
+      const tuix::Field *str = flatbuffers::GetTemporaryPointer(builder, child_offset);
+
+      if (str->value_type() != tuix::FieldUnion_StringField) {
+        throw std::runtime_error(
+          std::string("tuix::Upper requires str String, not ")
+          + std::string("str ")
+          + std::string(tuix::EnumNameFieldUnion(str->value_type())));
+      }
+
+      // Obtain the input as a string
+      bool result_is_null = str->is_null();
+
+      if (!result_is_null) {
+
+        auto str_field = static_cast<const tuix::StringField *>(str->value());
+
+        std::vector<uint8_t> str_vec(
+            flatbuffers::VectorIterator<uint8_t, uint8_t>(str_field->value()->Data(),
+                                                          static_cast<uint32_t>(0)),
+            flatbuffers::VectorIterator<uint8_t, uint8_t>(str_field->value()->Data(),
+                                                          static_cast<uint32_t>(str_field->length())));
+
+        std::string lower(str_vec.begin(), str_vec.end());
+
+        // Turn lower into uppercase and revert to vector
+        std::transform(lower.begin(), lower.end(), lower.begin(), ::toupper);
+        std::vector<uint8_t> result(lower.begin(), lower.end());
+
+        // Writing the result
+        return tuix::CreateField(
+          builder,
+          tuix::FieldUnion_StringField,
+          tuix::CreateStringFieldDirect(builder, &result, str_field->length()).Union(),
+          result_is_null);
+      } else {
+
+        // Creation is failing with null pointer. Trivially create empty string
+        std::string empty("\0");
+        std::vector<uint8_t> result(empty.begin(), empty.end());
+
+        return tuix::CreateField(
+          builder,
+          tuix::FieldUnion_StringField,
+          tuix::CreateStringFieldDirect(builder, &result, 0).Union(),
+          result_is_null);
+      }
+    }
+
     // Conditional expressions
     case tuix::ExprUnion_If:
     {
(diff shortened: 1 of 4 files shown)


docker/README.md
```diff
@@ -1,20 +0,0 @@
-Docker configuration for testing and development of Opaque.
```
wzheng, line 1: This diff shouldn't be here  since it's compared to the current master.

src/enclave/Enclave/ExpressionEvaluation.h
```diff
@@ -961,6 +961,40 @@ class FlatbuffersExpressionEvaluator {
...
+
+      if (str->value_type() != tuix::FieldUnion_StringField) {
+        throw std::runtime_error(
+          std::string("tuix::Substring requires str String, not ")
+          + std::string("str ")
+          + std::string(tuix::EnumNameFieldUnion(str->value_type())));
+      }
+
+      // Obtain the input as a string
+      bool result_is_null = str->is_null();
```
wzheng, line 977: If the input field is NULL, then no further processing needs to be done and we can just return a NULL value.

src/main/scala/edu/berkeley/cs/rise/opaque/expressions/Upper.scala
```diff
@@ -0,0 +1,33 @@
...
+
+case class Upper(child: Expression)
+    extends UnaryExpression with String2StringExpression with NullIntolerant {
+
+  override def convert(v: UTF8String): UTF8String = v.toUpperCase
+
+  override def doGenCode(ctx: CodegenContext, ev: ExprCode): ExprCode = {
+    defineCodeGen(ctx, ev, c => s"($c).toUpperCase()")
+  }
+}
```
wzheng, line 33: This file doesn't need to be here since Upper is already defined in Catalyst.

src/test/scala/edu/berkeley/cs/rise/opaque/OpaqueOperatorTests.scala
```diff
@@ -582,6 +589,20 @@ trait OpaqueOperatorTests extends FunSuite with BeforeAndAfterAll { self =>
     df.select(dot($"v1", $"v2")).collect
   }
 
+  testAgainstSpark("upper") { securityLevel =>
```
wzheng, line 592: I'm curious how this test passes... could you paste the logical and physical plan queries here?

src/main/scala/edu/berkeley/cs/rise/opaque/Utils.scala
```diff
@@ -106,6 +106,7 @@ import edu.berkeley.cs.rise.opaque.expressions.DotProduct
 import edu.berkeley.cs.rise.opaque.expressions.VectorAdd
 import edu.berkeley.cs.rise.opaque.expressions.VectorMultiply
 import edu.berkeley.cs.rise.opaque.expressions.VectorSum
+import edu.berkeley.cs.rise.opaque.expressions.Upper
```
wzheng, line 109: We should use Catalyst's Upper expression since it already exists. Upper shouldn't be a UDF.

wzheng (commented):
Partial review.

src/test/scala/edu/berkeley/cs/rise/opaque/OpaqueOperatorTests.scala
wzheng: Also, the test should test for NULL values as well.

eric-feng-2011: scala> dfe.select(upper($"v1")).explain
== Physical Plan ==
EncryptedProject [upper(v1#4) AS upper(v1)#26]
+- EncryptedLocalTableScan [v1#4, v2#5], [[lower,upper]]

src/main/scala/edu/berkeley/cs/rise/opaque/Utils.scala
eric-feng-2011: Changed to use catalyst upper

docker/README.md
eric-feng-2011: I think I got this resolved.

src/test/scala/edu/berkeley/cs/rise/opaque/OpaqueOperatorTests.scala
eric-feng-2011: Still working on adding the null test...

src/enclave/Enclave/ExpressionEvaluation.h
eric-feng-2011: Added null parsing. Doesn't seem to be working with null test...

src/test/scala/edu/berkeley/cs/rise/opaque/OpaqueOperatorTests.scala
chester-leung: Maybe you can create two tests, one for nulls and one for non-nulls

src/enclave/Enclave/ExpressionEvaluation.h
eric-feng-2011: Null test working. Null is null.asInstanceOf[String]

src/test/scala/edu/berkeley/cs/rise/opaque/OpaqueOperatorTests.scala
eric-feng-2011: Added two tests - one including null and one without

wzheng (approved):
Looks good, thanks!

PR #106, opened 2020-10-03
CaseWhen

andrewlawhh:
Add support for multiple branched CaseWhen expression and optional else/default value.

### DIFF
--- a/src/enclave/Enclave/ExpressionEvaluation.h
+++ b/src/enclave/Enclave/ExpressionEvaluation.h
@@ -742,6 +742,45 @@ private:
       }
     }
 
+    case tuix::ExprUnion_CaseWhen:
+    {
+      auto e = expr->expr_as_CaseWhen();
+      size_t num_children = e->children()->size();
+
+      // Evaluate to the first value whose predicate is true.
+      // Short circuit on the earliest branch possible.
+      for (size_t i = 0; i < num_children - 1; i += 2) {
+        auto predicate_offset = eval_helper(row, (*e->children())[i]);
+        auto true_value_offset = eval_helper(row, (*e->children())[i+1]);
+        const tuix::Field *predicate =
+          flatbuffers::GetTemporaryPointer(builder, predicate_offset);
+        const tuix::Field *true_value =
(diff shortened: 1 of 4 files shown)


src/test/scala/edu/berkeley/cs/rise/opaque/OpaqueOperatorTests.scala
```diff
@@ -395,6 +395,27 @@ trait OpaqueOperatorTests extends FunSuite with BeforeAndAfterAll { self =>
     df.select(year($"date")).collect
   }
 
+  testAgainstSpark("case when") { securityLevel =>
```
wzheng, line 398: I don't think this will test all of the queries, since `testAgainstSpark` will take the final returned values from this function and compare them.

src/enclave/Enclave/ExpressionEvaluation.h
```diff
@@ -742,6 +742,65 @@ class FlatbuffersExpressionEvaluator {
...
+        std::string empty("\0");
+        std::vector<uint8_t> result(empty.begin(), empty.end());
+        return tuix::CreateField(
+          builder,
+          tuix::FieldUnion_StringField,
+          tuix::CreateStringFieldDirect(builder, &result, 0).Union(),
+          true);
+      }
+      return tuix::CreateField(builder, result_type,
+          tuix::CreateNullField(builder).Union(), true);
```
wzheng, line 801: What underlying value will this give? It's a bit strange since the `result_type` will be something different, but the value passed to it is a `NullField`

andrewlawhh: I could be understanding wrong but I think .Union() returns Offset<void> regardless of the FieldType that calls it so it doesn't matter what I put there? Not entirely sure though.

I did that because the alternative is to have an enormous switch statement for every possible type that "result_type" could be, which is something we could do but is maybe verbose

wzheng: Ok, we can keep this for now. Can you put a comment here to explain your design decision?

PR #109, opened 2020-10-14
Add partition ID to enclave

chester-leung:
Computation Integrity PR 1/6

This portion of the PR adds the partition ID assigned by the driver to each partition (and the enclave). This partition ID is logged as part of the LogEntry and is used during post verification to ensure that blocks were shuffled properly.

This code assumes a trusted driver.


src/main/scala/edu/berkeley/cs/rise/opaque/execution/operators.scala
```diff
@@ -197,9 +200,10 @@ case class EncryptedProjectExec(projectList: Seq[NamedExpression], child: SparkP
   override def executeBlocked(): RDD[Block] = {
     val projectListSer = Utils.serializeProjectList(projectList, child.output)
     timeOperator(child.asInstanceOf[OpaqueOperatorExec].executeBlocked(), "EncryptedProjectExec") {
-      childRDD => childRDD.map { block =>
+      childRDD => 
+        childRDD.map { block =>
         val (enclave, eid) = Utils.initEnclave()
-        Block(enclave.Project(eid, projectListSer, block.bytes))
+        Block(enclave.Project(eid, projectListSer, block.bytes, TaskContext.getPartitionId))
```
wzheng, line 206: Why are you using `TaskContext.getPartitionId` here instead of `mapPartitionsWithIndex`?

Also, with either of those two choices, how does the worker enclave get the partition ID? Unless it's encrypted, I think the partition ID could also be changed by the server side's untrusted code, so we'd be relying on more than just the trusted driver.

PR #110, opened 2020-10-15
Like

andrewlawhh:


src/test/scala/edu/berkeley/cs/rise/opaque/OpaqueOperatorTests.scala
```diff
@@ -471,6 +471,42 @@ trait OpaqueOperatorTests extends FunSuite with BeforeAndAfterAll { self =>
     df.select(when(df("word") === "foo", 3).when(df("word") === "baz", 2)).collect
   }
 
+  testAgainstSpark("LIKE - Contains") { securityLevel =>
```
wzheng, line 474: Can you write the test so that the column being processed contains NULL values?

wzheng (commented):
Looks great, just one comment!

wzheng (approved):
Thanks!

PR #112, opened 2020-10-15
Support for LogEntry

chester-leung:
Computation Integrity PR 3/6

This PR adds support for creating and storing `LogEntry` objects in `EncryptedBlocks`. `LogEntry` objects store metadata of executed operators, and are used for both during-execution integrity checks as well as post verification.

This PR assumes a trusted driver.


src/enclave/Enclave/FlatbuffersWriters.cpp
```diff
@@ -75,30 +85,142 @@ void RowWriter::maybe_finish_block() {
...
+          le.job_id);
+      past_log_entries_vector.push_back(past_log_entry_serialized);
+    }
+
+    num_past_log_entries.push_back(past_log_entries.size());
+   
+    // We will MAC over global_mac || curr_ecall || snd_pid || rcv_pid || job_id || num_macs 
+    // || global_mac || num past log entries || past log entries
+    int num_past_entries = (int) past_log_entries.size();
+    int past_ecalls_lengths = get_past_ecalls_lengths(past_log_entries, 0, num_past_entries);
```
wzheng, line 190: If you're MAC'ing everything here, do you still need the specific global MAC per log entry?

src/enclave/Enclave/FlatbuffersWriters.cpp
```diff
@@ -75,30 +85,142 @@ void RowWriter::maybe_finish_block() {
...
+    //  * global_mac of size OE_HMAC_SIZE
+    //  * 5 ints
+    // 1 past log entry contains:
+    //  * string ecall of size ecall.length()
+    //  * 3 ints
+    int num_bytes_to_mac = OE_HMAC_SIZE + 5 * sizeof(int) + curr_ecall.length() * sizeof(char) 
+      + 3 * sizeof(int) * past_log_entries.size() + past_ecalls_lengths * sizeof(char);
+    uint8_t to_mac[num_bytes_to_mac];
+
+    uint8_t hmac[32];
```
wzheng, line 203: Isn't this more of a global MAC than the current global MAC in LogEntry? Maybe you can rename the `global_mac` to `log_mac`?

wzheng (commented):
Partial review.

src/main/scala/edu/berkeley/cs/rise/opaque/execution/operators.scala
```diff
@@ -243,14 +248,29 @@ case class EncryptedAggregateExec(
...
       // Send first row to previous partition and last group to next partition
-      val shiftedFirstRows = firstRows.drop(1) :+ Utils.emptyBlock
-      val shiftedLastGroups = Utils.emptyBlock +: lastGroups.dropRight(1)
-      val shiftedLastRows = Utils.emptyBlock +: lastRows.dropRight(1)
+      var shiftedFirstRows = Array[Block]()
+      var shiftedLastGroups = Array[Block]()
+      var shiftedLastRows = Array[Block]()
+      if (childRDD.getNumPartitions == 1) {
+        val firstRowDrop = firstRows(0)
+        shiftedFirstRows = firstRows.drop(1) :+ Utils.emptyBlock(firstRowDrop)
```
wzheng, line 261: You can just return the `emptyBlock` here?

src/enclave/Enclave/EnclaveContext.h
```diff
@@ -0,0 +1,215 @@
...
+    void append_past_log_entry(std::string ecall, int snd_pid, int rcv_pid, int job_id) {
+      LogEntry le;
+      le.ecall = ecall;
+      le.snd_pid = snd_pid;
+      le.rcv_pid = rcv_pid;
+      le.job_id = job_id;
+      ecall_log_entries.insert(le);
+    }
+
+    std::vector<LogEntry> get_ecall_log_entries() {
```
wzheng, line 112: Maybe rename this `get_past_log_entries`?

src/enclave/Enclave/FlatbuffersWriters.cpp
```diff
@@ -75,30 +85,142 @@ void RowWriter::maybe_finish_block() {
...
+    std::unique_ptr<uint8_t, decltype(&ocall_free)> mac_lst_ptr(untrusted_mac_lst, 
+        &ocall_free);
+    memcpy(mac_lst_ptr.get(), mac_lst, num_macs * SGX_AESGCM_MAC_SIZE);
+
+    // Copy global mac to untrusted memory
+    uint8_t* untrusted_global_mac = nullptr;
+    ocall_malloc(OE_HMAC_SIZE, &untrusted_global_mac);
+    std::unique_ptr<uint8_t, decltype(&ocall_free)> global_mac_ptr(untrusted_global_mac, 
+        &ocall_free);
+    memcpy(global_mac_ptr.get(), global_mac, OE_HMAC_SIZE);
```
wzheng, line 159: Why is there an ocall to main memory, instead of using enclave memory?

src/enclave/Enclave/FlatbuffersWriters.cpp
```diff
@@ -75,30 +85,142 @@ void RowWriter::maybe_finish_block() {
...
     finish_block();
   }
-  auto result = tuix::CreateEncryptedBlocksDirect(enc_block_builder, &enc_block_vector);
+
+  std::vector<flatbuffers::Offset<tuix::LogEntry>> curr_log_entry_vector;
+  std::vector<flatbuffers::Offset<tuix::LogEntry>> past_log_entries_vector;
+  std::vector<int> num_past_log_entries;
+  std::vector<flatbuffers::Offset<tuix::LogEntryChainMac>> log_entry_chain_hash_vector;
+  
+  if (curr_ecall != std::string("NULL")) {
```
wzheng, line 135: We could also using the empty string here, instead of using "NULL".

src/enclave/Enclave/EnclaveContext.h
```diff
@@ -0,0 +1,215 @@
...
+      std::vector<LogEntry> ret(ecall_log_entries.begin(), ecall_log_entries.end());
+      return ret;
+    }
```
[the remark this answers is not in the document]
chester-leung: `ecall_log_entries` is an unordered set, but I think there's no need to convert it to a vector here. I'll change this.

src/enclave/Enclave/Join.cpp
```diff
@@ -92,11 +96,13 @@ void non_oblivious_sort_merge_join(
...
 
+          EnclaveContext::getInstance().set_append_mac(true);
           w.append(primary, current);
         }
       }
     }
   }
 
-  w.output_buffer(output_rows, output_rows_length);
+  EnclaveContext::getInstance().set_append_mac(true);
```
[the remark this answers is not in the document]
chester-leung: `append_mac` is default true, so unless we at some point in aggregation set it to be false, we won't have to call this.

We do set `append_mac` to false at times during because we may call `output_buffer()` on blocks that aren't the result of the ecall, i.e. they're intermediate results of the operator and will not be received as input by a subsequent operator. We don't do integrity checks on these for now, but we should likely add this in later. (I made a note about this in the multiparty timeline doc)

src/enclave/Enclave/Join.cpp
```diff
@@ -61,11 +61,12 @@ void non_oblivious_sort_merge_join(
...
   while (r.has_next()) {
     const tuix::Row *current = r.next();
-
     if (join_expr_eval.is_primary(current)) {
+      EnclaveContext::getInstance().set_append_mac(false);
+      // If current row is from primary table
       if (last_primary_of_group.get()
           && join_expr_eval.is_same_group(last_primary_of_group.get(), current)) {
-        // Add this primary row to the current group
+        // Add this row to the current group
```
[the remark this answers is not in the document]
chester-leung: fixed

src/main/scala/edu/berkeley/cs/rise/opaque/execution/operators.scala
chester-leung: Right -  the result of the `firstRows.drop(1)` call (and the corresponding variables for shiftedLastGroups and shiftedLastRows) is essentially the empty block, but while the `Utils.emptyBlock(block)` function stores nothing in the `EncryptedBlocks` attribute, it does store some LogEntry data. So the block here is not exactly the `emptyBlock`. I coded it like this to make everything easier to read, but can remove the `drop(1)` and `dropRight` calls if you think that's better.

src/enclave/Enclave/FlatbuffersWriters.cpp
chester-leung: I've renamed this to `mac_lst_mac` as it's not exactly a MAC over the log entry.

src/enclave/Enclave/EnclaveContext.h
```diff
@@ -0,0 +1,215 @@
...
+
+    // For this ecall log entry
+    std::string this_ecall;
+    std::vector<std::vector<uint8_t>> log_entry_mac_lst;
+
+    std::string curr_row_writer;
+    // Special vectors of nonObliviousAggregateStep1
+    std::vector<std::vector<uint8_t>> first_row_log_entry_mac_lst;
+    std::vector<std::vector<uint8_t>> last_group_log_entry_mac_lst;
+    std::vector<std::vector<uint8_t>> last_row_log_entry_mac_lst;
```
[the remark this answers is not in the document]
chester-leung: The updates I've made don't include changes for this yet as it's more complicated, but I'll address this in later changes.

chester-leung:
I addressed most of these comments (except the one about hardcoded mac lists) in #122

PR #113, opened 2020-10-15
During execution integrity checks

chester-leung:
Computation Integrity PR 4/6

Adds support for during query execution integrity checks. In particular, we check that no EncryptedBlock's were added, dropped, or modified.


src/enclave/Enclave/IntegrityUtils.cpp
```diff
@@ -0,0 +1,172 @@
...
+  // Master list of mac lists of all input partitions
+  std::vector<std::vector<std::vector<uint8_t>>> partition_mac_lsts;
+
+  // Check that each input partition's global_mac is indeed a HMAC over the mac_lst
+  for (uint32_t i = 0; i < curr_entries_vec->size(); i++) {
+    auto input_log_entry = curr_entries_vec->Get(i);
+
+    // Copy over the global mac for this input log entry
+    uint8_t global_mac[SGX_AESGCM_MAC_SIZE];
+    memcpy(global_mac, input_log_entry->global_mac()->data(), SGX_AESGCM_MAC_SIZE);
```
wzheng, line 42: Why is this memcpy necessary? Can you access the deserialized flatbuffers data directly?

src/enclave/Enclave/IntegrityUtils.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+    // Copy over the mac_lst
+    int num_macs = input_log_entry->num_macs();
+    uint8_t mac_lst[num_macs * SGX_AESGCM_MAC_SIZE];
```
wzheng, line 46: Not sure we want to use VLAs here: https://stackoverflow.com/questions/1887097/why-arent-variable-length-arrays-part-of-the-c-standard

src/enclave/Enclave/IntegrityUtils.cpp
chester-leung: Fixed

src/enclave/Enclave/IntegrityUtils.cpp
chester-leung: Fixed. I'm now directly accessing the deserialized flatbuffers mac_lst, so no need for this as a destination buffer 
for `memcpy`

chester-leung:
These comments have been addressed in #123

PR #114, opened 2020-10-15
Integrity post verification

chester-leung:
Computation Integrity PR 5/6

This PR adds post verification of query execution. We check that data was shuffled properly and that operators were executed in the proper order according to the query plan.

This assumes a trusted driver (client/driver are coupled and the JobVerificationEngine is run by the driver), that all ecalls are run at every partition, and that each dataframe has the same number of partitions.


src/main/scala/edu/berkeley/cs/rise/opaque/execution/EncryptedSortExec.scala
```diff
@@ -41,37 +43,39 @@ object EncryptedSortExec {
     Utils.ensureCached(childRDD)
     time("force child of EncryptedSort") { childRDD.count }
     // RA.initRA(childRDD)
+    JobVerificationEngine.addExpectedOperator("EncryptedSortExec")
```
wzheng, line 46: Can this be place somewhere else? Given a DataFrame, you can get the physical plan directly by calling `df.queryExecution`: https://stackoverflow.com/questions/41716295/what-is-the-use-of-queryexecution-in-spark-dataframe

PR #115, opened 2020-10-15
Test modifications to support integrity implementation

chester-leung:
Computation Integrity PR 6/6

Modifies tests to enable integrity. We drop `Utils.time()` calls as this forces the query to run on this line, meaning that certain parts of the query are run multiple times by the end of the test, which messes up the integrity post verification.

We also comment out the "join with different number of partitions" test since integrity currently doesn't support dataframes with different number of partitions.


src/main/scala/edu/berkeley/cs/rise/opaque/benchmark/BigDataBenchmark.scala
```diff
@@ -58,14 +58,14 @@ object BigDataBenchmark {
...
-    Utils.time("load rankings") { Utils.force(rankingsDF) }
+    // Utils.time("load rankings") { Utils.force(rankingsDF) }
     Utils.timeBenchmark(
       "distributed" -> (numPartitions > 1),
       "query" -> "big data 1",
       "system" -> securityLevel.name,
       "size" -> size) {
       val df = rankingsDF.filter($"pageRank" > 1000)
-      Utils.force(df)
+      // Utils.force(df)
```
wzheng, line 68: Why was this commented out?

PR #124, opened 2020-11-18
Separate IN PR

Chenyu-Shi:
Split the IN&CONCAT pr by deleting all CONCAT changes

### DIFF
--- a/src/enclave/Enclave/ExpressionEvaluation.h
+++ b/src/enclave/Enclave/ExpressionEvaluation.h
@@ -738,6 +738,56 @@ private:
 
     }
 
+    case tuix::ExprUnion_Concat:
+    {
+      //implementing this like string concat 
+      auto c = static_cast<const tuix::Concat *>(expr->expr());
+      size_t num_children = c->children()->size(); 
+      //TODO add type checking 
+      size_t total = 0; 
+
+      std::vector<uint8_t> result; 
+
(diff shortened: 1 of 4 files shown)


src/enclave/Enclave/ExpressionEvaluation.h
```diff
@@ -702,6 +702,59 @@ class FlatbuffersExpressionEvaluator {
...
+        const tuix::Field *item = flatbuffers::GetTemporaryPointer(builder, right_offset);
+        if (item->value_type() != left->value_type()){
+          throw std::runtime_error(
+          std::string("In can't operate on ")
+          + std::string(tuix::EnumNameFieldUnion(left->value_type()))
+          + std::string(" and ")
+          + std::string(tuix::EnumNameFieldUnion(item->value_type()))
+          + ". Please double check the type of each input");
+        }
+        result_is_null = result_is_null || item ->is_null();
```
wzheng, line 732: Will this set the result to be NULL if any item is NULL? Even if there is a match after the NULL value is iterated over?

Chenyu-Shi: fixed. Now it's only null when there's no match and there are null inputs

wzheng:
Great, can you resolve the conflicts, and then we can merge it?

wzheng (approved):

PR #125, opened 2020-11-18
Separate Concat PR 

Chenyu-Shi:
Split the IN&CONCAT pr by deleting all IN changes

### DIFF
--- a/src/enclave/Enclave/ExpressionEvaluation.h
+++ b/src/enclave/Enclave/ExpressionEvaluation.h
@@ -738,6 +738,56 @@ private:
 
     }
 
+    case tuix::ExprUnion_Concat:
+    {
+      //implementing this like string concat 
+      auto c = static_cast<const tuix::Concat *>(expr->expr());
+      size_t num_children = c->children()->size(); 
+      //TODO add type checking 
+      size_t total = 0; 
+
+      std::vector<uint8_t> result; 
+
+      for (size_t i =0; i< num_children; i++){
+              auto offset =  eval_helper(row, (*c->children())[i]);
+              const tuix::Field *str = flatbuffers::GetTemporaryPointer(builder, offset);
+              if (str->value_type() != tuix::FieldUnion_StringField) {
+                throw std::runtime_error(
+                  std::string("tuix::Concat requires String, not ")
+                  + std::string(tuix::EnumNameFieldUnion(str->value_type())));  
+              }
+              auto str_field = static_cast<const tuix::StringField *>(str->value());
+              int32_t start = 0;
+              int32_t end = str_field ->length(); 
+              total += end; 
+
+              std::vector<uint8_t> stringtoadd(
+              flatbuffers::VectorIterator<uint8_t, uint8_t>(str_field->value()->Data(),
+                                                        static_cast<uint32_t>(start)),
+              flatbuffers::VectorIterator<uint8_t, uint8_t>(str_field->value()->Data(),
+                                                        static_cast<uint32_t>(end)));
+              result.insert(result.end(), stringtoadd.begin(), stringtoadd.end());
+      }
+      //TODO add null check 
+
+      return tuix::CreateField(
+          builder,
+          tuix::FieldUnion_StringField,
+          tuix::CreateStringFieldDirect(
+            builder, &result, static_cast<uint32_t>(total)).Union(),
+          false);
+      /*
+      auto array_field = static_cast<const tuix::ArrayField *>(value->value());
+      std::string str = to_string(array_field);
+      std::vector<uint8_t> str_vec(str.begin(), str.end());
+        return tuix::CreateField(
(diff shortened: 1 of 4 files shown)


src/enclave/Enclave/ExpressionEvaluation.h
```diff
@@ -702,6 +702,49 @@ class FlatbuffersExpressionEvaluator {
...
+                  std::string("tuix::Concat requires string, not ")
+                  + std::string(tuix::EnumNameFieldUnion(str->value_type()))
+                  + std::string(". You do not need to provide the data as string but the data should be serialized into string before sent to concat"));  
+              }
+              if (!str->is_null()){
+                // skipping over the null input 
+                auto str_field = static_cast<const tuix::StringField *>(str->value());
+                int32_t start = 0;
+                int32_t end = str_field ->length(); 
+                total += end;
```
wzheng, line 729: I think this will return a NULL value when concatenating with an empty string? Can you first add another test to test empty strings, then modify it if that's the case?

src/test/scala/edu/berkeley/cs/rise/opaque/OpaqueOperatorTests.scala
```diff
@@ -389,6 +389,22 @@ trait OpaqueOperatorTests extends FunSuite with BeforeAndAfterAll { self =>
     df.filter($"word".contains(lit("1"))).collect
   }
 
+  testAgainstSpark("concat with string") { securityLevel =>
```
wzheng, line 392: Add a test case for empty string.

src/enclave/Enclave/ExpressionEvaluation.h
```diff
@@ -702,6 +702,49 @@ class FlatbuffersExpressionEvaluator {
...
+                throw std::runtime_error(
+                  std::string("tuix::Concat requires string, not ")
+                  + std::string(tuix::EnumNameFieldUnion(str->value_type()))
+                  + std::string(". You do not need to provide the data as string but the data should be serialized into string before sent to concat"));  
+              }
+              if (!str->is_null()){
+                // skipping over the null input 
+                auto str_field = static_cast<const tuix::StringField *>(str->value());
+                int32_t start = 0;
+                int32_t end = str_field ->length();
```
wzheng, line 728: Why are you using int32_t here, and then casting them to uint32_t later?

src/enclave/Enclave/ExpressionEvaluation.h
```diff
@@ -702,6 +702,49 @@ class FlatbuffersExpressionEvaluator {
...
+      std::vector<uint8_t> result; 
+
+      for (size_t i =0; i< num_children; i++){
+              auto offset =  eval_helper(row, (*c->children())[i]);
+              const tuix::Field *str = flatbuffers::GetTemporaryPointer(builder, offset);
+              if (str->value_type() != tuix::FieldUnion_StringField) {
+                throw std::runtime_error(
+                  std::string("tuix::Concat requires string, not ")
+                  + std::string(tuix::EnumNameFieldUnion(str->value_type()))
+                  + std::string(". You do not need to provide the data as string but the data should be serialized into string before sent to concat"));
```
wzheng, line 722: I'm a bit confused about this runtime error's message -- seems like we just do not allowing data of another type, since a runtime error is thrown?

src/flatbuffers/Expr.fbs
```diff
@@ -118,6 +119,15 @@ table Contains {
     right:Expr;
 }
 
+// Array expressions
+table In{
+    children:[Expr]; 
+}
```
wzheng, line 125: This definition shouldn't appear in this PR, since it's already in the IN PR.

src/enclave/Enclave/ExpressionEvaluation.h
Chenyu-Shi: Good point. I added an empty string test in addition to the null test and seems like concat just Ignores both

src/test/scala/edu/berkeley/cs/rise/opaque/OpaqueOperatorTests.scala
Chenyu-Shi: combined with "concat with other datatype test"

src/enclave/Enclave/ExpressionEvaluation.h
Chenyu-Shi: fixed. Good point here

src/enclave/Enclave/ExpressionEvaluation.h
Chenyu-Shi: Changed the runtime error to better reflect what needs to happen

src/flatbuffers/Expr.fbs
Chenyu-Shi: Removed

PR #129, opened 2020-12-03
Changing operator matching from logical to physical

wzheng:
This PR transitions Opaque operator matching from the logical level to the physical level. This change is important in supporting subqueries in the future because it allows Spark to apply logical rules for such queries.


chester-leung (approved):

chester-leung:
LGTM!

Issue #117, opened 2020-11-05
Global aggregation returns incorrect results when the input to the aggregation has 0 rows


wzheng:
Currently, global aggregation works correctly when the input is non-empty. If the input is empty, then Opaque will return empty, but Spark SQL will return a result (e.g., return 0 for `count(*)`).

PR #132, opened 2020-12-19
Aggregation rewrite

wzheng:
This PR implements a new version of aggregation that no longer relies on boundary processing, which is an old artifact from the oblivious implementation. This new version follows the Spark aggregation and uses "partial" and "final" modes to aggregate. The "complete" mode is currently implemented but not used, though we should look into utilizing it for high-cardinality aggregation in the future. 

Features: 
- Aggregation has different modes: partial, final, and complete
- Sort now has a new parameter that indicates whether it's a global sort or a local per-partition sort
- Fixes issue #117

### DIFF
--- a/src/enclave/Enclave/Aggregate.cpp
+++ b/src/enclave/Enclave/Aggregate.cpp
@@ -5,12 +5,11 @@
 #include "FlatbuffersWriters.h"
 #include "common.h"
 
-
-void non_oblivious_partial_aggregate(
-  bool is_partial,
+void non_oblivious_aggregate(
   uint8_t *agg_op, size_t agg_op_length,
   uint8_t *input_rows, size_t input_rows_length,
-  uint8_t **partial_aggregates, size_t *partial_aggregates_length) {
+  uint8_t **output_rows, size_t *output_rows_length,
+  bool is_partial) {
 
   FlatbuffersAggOpEvaluator agg_op_eval(agg_op, agg_op_length);
   RowReader r(BufferRefView<tuix::EncryptedBlocks>(input_rows, input_rows_length));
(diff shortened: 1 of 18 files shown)


src/enclave/Enclave/ExpressionEvaluation.h
```diff
@@ -1446,6 +1609,10 @@ class FlatbuffersAggOpEvaluator {
     reset_group();
   }
 
+  size_t get_num_groups() {
```
ankurdave, line 1612: At the call site, "num_groups" sounds like it's referring to the number of distinct groups in the input. How about `get_num_grouping_keys()` or `get_num_grouping_exprs()`?

src/main/scala/edu/berkeley/cs/rise/opaque/Utils.scala
```diff
@@ -1112,22 +1128,25 @@ object Utils extends Logging {
   }
 
   def serializeAggOp(
-    groupingExpressions: Seq[Expression],
-    aggExpressions: Seq[NamedExpression],
+    groupingExpressions: Seq[NamedExpression],
+    aggExpressions: Seq[AggregateExpression],
     input: Seq[Attribute]): Array[Byte] = {
-    // aggExpressions contains both grouping expressions and AggregateExpressions. Transform the
-    // grouping expressions into AggregateExpressions that collect the first seen value.
```
ankurdave, line 1119: Maybe it's helpful to preserve the part of this comment that's still relevant, and perhaps augment it with some more context? Something like "The output of agg operator contains both the grouping columns and the aggregate values. To avoid the need for special handling of the grouping columns, we transform the grouping expressions into AggregateExpressions that collect the first seen value."

src/enclave/Enclave/Aggregate.cpp
```diff
@@ -5,116 +5,38 @@
...
-    }
-    agg_op_eval.aggregate(cur.get());
-
-    // Output the current aggregate if it is the last aggregate for its run
-    if (next.get() == nullptr || !agg_op_eval.is_same_group(cur.get(), next.get())) {
-      w.append(agg_op_eval.evaluate());
-    }
+  // Skip outputting the final row if the number of input rows is 0 AND
+  // 1. It's a grouping aggregation, OR
+  // 2. It's a global aggregation, the mode is final
```
ankurdave, line 35: Do we need to skip outputting the final row in case 2? If I understand correctly, that can only occur if there are zero partitions in the partial stage. Not sure if that's possible, but if it is, global aggregation should still return a row.

If we don't need to skip in case 2, it would greatly simplify the condition and comment.

src/main/scala/edu/berkeley/cs/rise/opaque/strategies.scala
```diff
@@ -19,69 +19,138 @@ package edu.berkeley.cs.rise.opaque
...
+      } else {
+        // Grouping aggregation
+        EncryptedProjectExec(resultExpressions,
+          EncryptedAggregateExec(
+            groupingExpressions, aggregateExpressions, Final,
+            EncryptedSortExec(groupingExpressions.map(_.toAttribute).map(e => SortOrder(e, Ascending)), true,
+              EncryptedAggregateExec(
+                groupingExpressions, aggregateExpressions, Partial, 
+                EncryptedSortExec(
+                  groupingExpressions.map(e => SortOrder(e, Ascending)), false, planLater(child)))))) :: Nil
```
ankurdave, line 120: Great that we no longer shuffle the full unaggregated input!

ankurdave (commented):
Looks like a great change! I have minor comments.

src/enclave/Enclave/Aggregate.cpp
wzheng: If the second statement isn't added, then the global aggregation test fails when there are multiple partitions, since it will output a row for every empty partition as well:

```[info] - global aggregate - encrypted *** FAILED *** (433 milliseconds)
[info]   Array([0], [256], [0]) did not equal Array([256]) (OpaqueOperatorTests.scala:90)```

File: src/main/scala/edu/berkeley/cs/rise/opaque/Utils.scala (hunk 2 above)
Comment (wzheng) [reply src/main/scala/edu/berkeley/cs/rise/opaque/Utils.scala]: Added, thanks!

**Review by chester-leung — COMMENTED** [review]
Looks good! Just a couple nits.

File: src/enclave/Enclave/Aggregate.cpp [hunk 5]
```diff
@@ -5,116 +5,40 @@
...
+void non_oblivious_aggregate(
   uint8_t *agg_op, size_t agg_op_length,
   uint8_t *input_rows, size_t input_rows_length,
-  uint8_t **first_row, size_t *first_row_length,
-  uint8_t **last_group, size_t *last_group_length,
-  uint8_t **last_row, size_t *last_row_length) {
+  uint8_t **output_rows, size_t *output_rows_length,
+  bool is_partial) {
+
+  (void) is_partial;
```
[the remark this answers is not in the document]
wzheng: Good catch, this is unnecessary -- `is_partial` is used in the last part of this function.

chester-leung (approved):

PR #136, opened 2021-01-26
TPC-H test suite added

octaviansima:
This PR refactors the single existing TPC-H query to perform operations using standard SQL rather than directly with DataFrames. This was done to expedite the process of increasing TPC-H coverage, since all of the queries are already available in SQL. It also creates a new folder to hold all queries from https://github.com/apache/spark/tree/master/sql/core/src/test/resources/tpch

### DIFF
--- a/src/main/scala/edu/berkeley/cs/rise/opaque/benchmark/TPCH.scala
+++ b/src/main/scala/edu/berkeley/cs/rise/opaque/benchmark/TPCH.scala
@@ -174,6 +174,20 @@ object TPCHDataFrames {
       .option("delimiter", "|")
       .load(s"${Benchmark.dataDir}/tpch/$size/customer.tbl")
       .repartition(numPartitions))
+
+  def generateMap(
+      sqlContext: SQLContext, securityLevel: SecurityLevel, size: String, numPartitions: Int)
+      : Map[String, DataFrame] = {
+    Map("part" -> TPCHDataFrames.part(sqlContext, securityLevel, size, numPartitions),
+    "supplier" -> TPCHDataFrames.supplier(sqlContext, securityLevel, size, numPartitions),
+    "lineitem" -> TPCHDataFrames.lineitem(sqlContext, securityLevel, size, numPartitions),
+    "partsupp" -> TPCHDataFrames.partsupp(sqlContext, securityLevel, size, numPartitions),
+    "orders" -> TPCHDataFrames.orders(sqlContext, securityLevel, size, numPartitions),
+    "nation" -> TPCHDataFrames.nation(sqlContext, securityLevel, size, numPartitions),
+    "region" -> TPCHDataFrames.region(sqlContext, securityLevel, size, numPartitions),
+    "customer" -> TPCHDataFrames.customer(sqlContext, securityLevel, size, numPartitions)
+    ),
+  }
 }
 
 class TPCH(
@@ -185,7 +199,7 @@ class TPCH(
   val nameToEncryptedDF: Map[String, DataFrame],
 ) {
 
-  val tableNames = Seq("part", "supplier", "lineitem", "partsupp", "orders", "nation")
+  val tableNames = Seq("part", "supplier", "lineitem", "partsupp", "orders", "nation", "region", "customer")
 
   def this(
     sqlContext: SQLContext,
@@ -193,22 +207,8 @@ class TPCH(
     numPartitions: Int,
   ) = {
     this(sqlContext, size, numPartitions, 
-
-    Map("part" -> TPCHDataFrames.part(sqlContext, Insecure, size, numPartitions),
-    "supplier" -> TPCHDataFrames.supplier(sqlContext, Insecure, size, numPartitions),
-    "lineitem" -> TPCHDataFrames.lineitem(sqlContext, Insecure, size, numPartitions),
-    "partsupp" -> TPCHDataFrames.partsupp(sqlContext, Insecure, size, numPartitions),
-    "orders" -> TPCHDataFrames.orders(sqlContext, Insecure, size, numPartitions),
-    "nation" -> TPCHDataFrames.nation(sqlContext, Insecure, size, numPartitions)),
-
-    Map("part" -> TPCHDataFrames.part(sqlContext, Encrypted, size, numPartitions),
-    "supplier" -> TPCHDataFrames.supplier(sqlContext, Encrypted, size, numPartitions),
-    "lineitem" -> TPCHDataFrames.lineitem(sqlContext, Encrypted, size, numPartitions),
-    "partsupp" -> TPCHDataFrames.partsupp(sqlContext, Encrypted, size, numPartitions),
-    "orders" -> TPCHDataFrames.orders(sqlContext, Encrypted, size, numPartitions),
-    "nation" -> TPCHDataFrames.nation(sqlContext, Encrypted, size, numPartitions)),
-    )
-    ensureCached()
+        TPCHDataFrames.generateMap(sqlContext, Insecure, size, numPartitions),
+        TPCHDataFrames.generateMap(sqlContext, Encrypted, size, numPartitions))
   }
 
   def ensureCached() = {
@@ -222,10 +222,7 @@ class TPCH(
     }
   }
 
-  def query(queryNumber: Int, securityLevel: SecurityLevel, sqlContext: SQLContext) : DataFrame = {
-    val queryLocation = sys.env.getOrElse("OPAQUE_HOME", ".") + "/src/test/resources/tpch/"
-    val sqlStr = Source.fromFile(queryLocation + s"q$queryNumber.sql").getLines().mkString("\n")
-
+  def setupViews(securityLevel: SecurityLevel) = {
     securityLevel match {
       case Insecure => {
         for ((name, df) <- nameToDF) {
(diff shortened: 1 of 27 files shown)


src/main/scala/edu/berkeley/cs/rise/opaque/benchmark/TPCH.scala
```diff
@@ -17,15 +17,19 @@
...
 
+import java.io.File
+import scala.io.Source
+
 import org.apache.spark.sql.DataFrame
-import org.apache.spark.sql.functions._
 import org.apache.spark.sql.types._
 import org.apache.spark.sql.SQLContext
+import org.apache.spark.sql.catalyst.util.fileToString
+import org.apache.spark.sql.catalyst.util.resourceToString
```
wzheng, line 27: Are these used?

src/main/scala/edu/berkeley/cs/rise/opaque/benchmark/TPCH.scala
```diff
@@ -17,15 +17,19 @@
...
 
 object TPCH {
   def part(
       sqlContext: SQLContext, securityLevel: SecurityLevel, size: String, numPartitions: Int)
-      : DataFrame =
+      : Unit =
```
wzheng, line 32: I don't think you have to put `Unit` here? Won't `def part(xx) = {}` work?

wzheng (commented):
Looks good, just a couple of comments!

src/main/scala/edu/berkeley/cs/rise/opaque/benchmark/TPCH.scala
```diff
@@ -138,59 +141,71 @@ object TPCH {
...
+
+    Map("part" -> TPCHDataFrames.part(sqlContext, Encrypted, size, numPartitions),
+    "supplier" -> TPCHDataFrames.supplier(sqlContext, Encrypted, size, numPartitions),
+    "lineitem" -> TPCHDataFrames.lineitem(sqlContext, Encrypted, size, numPartitions),
+    "partsupp" -> TPCHDataFrames.partsupp(sqlContext, Encrypted, size, numPartitions),
+    "orders" -> TPCHDataFrames.orders(sqlContext, Encrypted, size, numPartitions),
+    "nation" -> TPCHDataFrames.nation(sqlContext, Encrypted, size, numPartitions)),
+    )
+    ensureCached()
+  }
```
wzheng, line 179: The way that this secondary constructor written is a bit confusing. I wonder if you can have an `init()` to set up this mapping and ensure that the DataFrames are cached when `TPCHTests` is called.

src/main/scala/edu/berkeley/cs/rise/opaque/benchmark/TPCH.scala
```diff
@@ -138,59 +141,71 @@ object TPCH {
...
+        for ((name, df) <- nameToDF) {
+          df.createOrReplaceTempView(name)
+        }
+      }
+      case Encrypted => {
+        for ((name, df) <- nameToEncryptedDF) {
+          df.createOrReplaceTempView(name)
+        }
+      }
+    }
```
wzheng, line 207: Does `createOrReplaceTempView` need to be called every time there is a query? Or do these views last for a single Spark session? If so, then the SQL query may be using the wrong DataFrame. Can you double check by printing out the query's plan, using `df.explain()`? (`df.explain(true)` is also useful and gives both logical and physical plans).

src/main/scala/edu/berkeley/cs/rise/opaque/benchmark/TPCH.scala
wzheng, line 179: This constructor setup is currently a bit confusing to read. Is it possible to put this in a separate setup function, which can then be called by TPCHTests before the tests are run?

src/main/scala/edu/berkeley/cs/rise/opaque/benchmark/TPCH.scala
wzheng, line 207: Can the view creation be put in the setup function as well?

src/test/scala/edu/berkeley/cs/rise/opaque/TPCHTests.scala
```diff
@@ -0,0 +1,60 @@
...
+  testAgainstSpark("TPC-H 9") { securityLevel =>
+    tpch.query(9, securityLevel, spark.sqlContext).collect.toSet
+  }
+}
+
+class TPCHSinglePartitionSuite extends TPCHTests {
+  override def numPartitions: Int = 1
+  override val spark = SparkSession.builder()
+    .master("local[1]")
+    .appName("QEDSuite")
```
wzheng, line 44: Small nit: I think this appName can be changed to something like `TPCHSuite`. The same goes for OpaqueOperatorTests. This naming seems to have been carried over from the very first test suite we had (QEDSuite).

src/test/scala/edu/berkeley/cs/rise/opaque/TPCHTests.scala
```diff
@@ -0,0 +1,65 @@
...
+
+trait TPCHTests extends OpaqueTestsBase { self => 
+
+  def size = "sf_small"
+  def tpch: TPCH
+
+  override def beforeAll(): Unit = {
+    super.beforeAll();
+    tpch.ensureCached();
+  }
```
wzheng, line 34: This is great, thanks! Can you also call `generateMap` in here? Or put both `ensureCached` and `generateMap` in an init function.

src/test/scala/edu/berkeley/cs/rise/opaque/TPCHTests.scala
```diff
@@ -0,0 +1,65 @@
...
+    .getOrCreate()
+
+  override def tpch = new TPCH(spark.sqlContext, size, numPartitions)
+}
+
+class TPCHMultiplePartitionSuite extends TPCHTests {
+  override def numPartitions: Int = 3
+  override val spark = SparkSession.builder()
+    .master("local[1]")
+    .appName("QEDSuite")
```
wzheng, line 60: Can you change the app name here as well?

src/test/scala/edu/berkeley/cs/rise/opaque/TPCHTests.scala
```diff
@@ -0,0 +1,65 @@
...
+
+class TPCHSinglePartitionSuite extends TPCHTests {
+  override def numPartitions: Int = 1
+  override val spark = SparkSession.builder()
+    .master("local[1]")
+    .appName("QEDSuite")
+    .config("spark.sql.shuffle.partitions", numPartitions)
+    .getOrCreate()
+
+  override def tpch = new TPCH(spark.sqlContext, size, numPartitions)
```
wzheng, line 53: It seems like creating a new TPCH object for each test will create the DataFrames twice? Can you only load the tables once for both single partition and multi-partition tests?

src/test/scala/edu/berkeley/cs/rise/opaque/OpaqueOperatorTests.scala
```diff
@@ -911,7 +831,7 @@ trait OpaqueOperatorTests extends FunSuite with BeforeAndAfterAll { self =>
 
 }
 
-class OpaqueSinglePartitionSuite extends OpaqueOperatorTests {
+class OpaqueOperatorSinglePartitionSuite extends OpaqueOperatorTests {
   override val spark = SparkSession.builder()
     .master("local[1]")
     .appName("QEDSuite")
```
wzheng, line 837: Can you modify the name here?

src/test/scala/edu/berkeley/cs/rise/opaque/OpaqueOperatorTests.scala
```diff
@@ -921,7 +841,7 @@ class OpaqueSinglePartitionSuite extends OpaqueOperatorTests {
   override def numPartitions: Int = 1
 }
 
-class OpaqueMultiplePartitionSuite extends OpaqueOperatorTests {
+class OpaqueOperatorMultiplePartitionSuite extends OpaqueOperatorTests {
   override val spark = SparkSession.builder()
     .master("local[1]")
     .appName("QEDSuite")
```
wzheng, line 847: Modify the name here too?

wzheng (commented):
Just a few more comments!

src/test/scala/edu/berkeley/cs/rise/opaque/TPCHTests.scala
octaviansima: I can't because `generateMap` returns the actual map structure that the constructor needs. One alternative would be to do something like `val nameToDF = TPCHDataFrames.generateMap(sqlContext, Insecure, size, numPartitions)` outside of the auxiliary `this`constructor. The other is to set `nameToDF` to an arbitrary initialized value, then call an init function that sets it to the correct value while also calling `ensureCached`.

src/test/scala/edu/berkeley/cs/rise/opaque/TPCHTests.scala
octaviansima: I can't because to create the DataFrame you need a number of partitions. The two sets of DataFrames have the same schema and rows, but they are partitioned differently.

src/test/scala/edu/berkeley/cs/rise/opaque/TPCHTests.scala
wzheng: The current code is just a bit confusing, since the default constructor is not meant to be used directly. I think `val nameToDF = TPCHDataFrames.generateMap(sqlContext, Insecure, size, numPartitions)` is good. Perhaps another way is to use a companion object and define `apply ` methods.

octaviansima: I went with the `apply` method because that enables `generateMap` and `ensureCached` to be called in the same spot as a sort of init function. Thanks for the tip!

wzheng:
Thanks!

PR #139, opened 2021-01-29
Add TPC-H Benchmarks

octaviansima:
This PR adds support for the existing Benchmark.scala to run TPC-H queries with the additional specification of a command-line argument (or run all benchmarks if omitted). It also includes minor restructuring in TPCH.scala to allow for running `Utils.timeBenchmark` around only the query operator.


src/main/scala/edu/berkeley/cs/rise/opaque/benchmark/TPCHBenchmark.scala
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+package edu.berkeley.cs.rise.opaque.benchmark
+
+import edu.berkeley.cs.rise.opaque.Utils
+
+import org.apache.spark.sql.SQLContext
+
+object TPCHBenchmark {
+  def size = "sf_small"
```
wzheng, line 25: Maybe this should be a variable that can be set from the commandline -- I'm guessing we will most likely have more than one data size for testing in the future.

src/main/scala/edu/berkeley/cs/rise/opaque/benchmark/TPCHBenchmark.scala
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  def run(sqlContext: SQLContext) = {
+    val tpch = TPCH(sqlContext, size)
+
+    val supportedQueries = Seq(1, 3, 5, 6, 7, 8, 9, 10, 14, 17)
+
+    for (queryNumber <- supportedQueries) {
+      query(queryNumber, tpch, sqlContext, 1)
+      query(queryNumber, tpch, sqlContext, 3)
```
wzheng, line 56: Maybe change the number of partitions can be input from the commandline as well; or if unspecified, then by default be 2x the number of workers?

src/main/scala/edu/berkeley/cs/rise/opaque/benchmark/Benchmark.scala
```diff
@@ -24,35 +24,81 @@ import org.apache.spark.sql.SparkSession
...
+  def main(args: Array[String]): Unit = {
+    Utils.initSQLContext(spark.sqlContext)
 
+    var runAll = true
+    args.sliding(2, 2).toList.collect {
+      case Array("--num-partitions", numPartitions: String) => {
+        this.numPartitions = numPartitions.toInt
+      }
+      case Array("--size", size: String) => {
+        this.size = size
```
wzheng, line 82: Can you put some some error handling for when the size does not exist? Or give a list of sizes to the user?

src/main/scala/edu/berkeley/cs/rise/opaque/benchmark/Benchmark.scala
```diff
@@ -24,35 +24,81 @@ import org.apache.spark.sql.SparkSession
...
     LogisticRegression.train(spark, Insecure, 100000, 1)
     LogisticRegression.train(spark, Encrypted, 100000, 1)
+  }
+
+  def runAll() = {
+    logisticRegression()
+    TPCHBenchmark.run(spark.sqlContext, numPartitions, size)
+  }
+
+  def main(args: Array[String]): Unit = {
```
wzheng, line 73: I wonder if we should have a help argument here? Also, once this benchmark is set up, it might be good to put some text in the README about this benchmark.

src/main/scala/edu/berkeley/cs/rise/opaque/benchmark/Benchmark.scala
```diff
@@ -24,35 +24,81 @@ import org.apache.spark.sql.SparkSession
...
+        this.size = size
+      }
+      case Array("--operations", operations: String) => {
+        runAll = false
+        val operationsArr = operations.split(",").map(_.trim)
+        for (operation <- operationsArr) {
+          operation match {
+            case "logistic-regression" => {
+              logisticRegression()
+            }
```
wzheng, line 91: Did you make sure that this actually runs? I remember there were some errors that cause this to fail a while ago...

src/test/scala/edu/berkeley/cs/rise/opaque/TPCHTests.scala
```diff
@@ -26,94 +26,94 @@ import edu.berkeley.cs.rise.opaque.benchmark.TPCH
 trait TPCHTests extends OpaqueTestsBase { self => 
 
   def size = "sf_small"
-  def tpch = TPCH(spark.sqlContext, size)
+  def tpch = new TPCH(spark.sqlContext, size, "file://")
 
   testAgainstSpark("TPC-H 1") { securityLevel =>
-    tpch.query(1, securityLevel, spark.sqlContext, numPartitions).collect
+    tpch.query(1, securityLevel, numPartitions).collect
   }
```
wzheng, line 33: Maybe this can be simplified to have a function make a call repeated to testAgainstSpark for these query numbers?

PR #142, opened 2021-02-02
TPC-H 20 Fix

octaviansima:
This PR fixes the matching errors encountered by the TPC-H 20 test by having decimals be treated as floats in Utils.scala. This has to be done because decimals are not supported yet by the current Flatbuffer serialization.


src/main/scala/edu/berkeley/cs/rise/opaque/Utils.scala
```diff
@@ -403,6 +402,18 @@ object Utils extends Logging {
           tuix.FieldUnion.FloatField,
           tuix.FloatField.createFloatField(builder, 0),
           isNull)
+      case (x: Decimal, DecimalType()) =>
+        tuix.Field.createField(
+          builder,
+          tuix.FieldUnion.FloatField,
+          tuix.FloatField.createFloatField(builder, x.toString().toFloat),
```
wzheng, line 409: There seems to be a "toFloat" function for the Decimal type? Can you use that?

wzheng (commented):
Just one comment!

wzheng (approved):

PR #145, opened 2021-02-04
Join update

wzheng:
This PR has two contributions:
1. Adds support for two more join types: left semijoin and left antijoin. Currently also fixes TPC-H query 4.
2. Changes the implementation of join to no longer require passing the last primary group to from the previous partition to the next partition. This is done by splitting the sort operation into two separate operations: partition and sort. The new join operator partitions and sorts on different columns, thus eliminating the limitation specified in this [comment](https://github.com/mc2-project/opaque/blob/0d69b7be0cfc8df129392278cb9d4da4f69f1124/src/enclave/Enclave/Join.cpp#L21).

### DIFF
--- a/src/main/scala/edu/berkeley/cs/rise/opaque/strategies.scala
+++ b/src/main/scala/edu/berkeley/cs/rise/opaque/strategies.scala
@@ -33,6 +33,7 @@ import org.apache.spark.sql.catalyst.planning.ExtractEquiJoinKeys
 import org.apache.spark.sql.catalyst.planning.PhysicalAggregation
 import org.apache.spark.sql.catalyst.plans.logical._
 import org.apache.spark.sql.catalyst.plans.Inner
+import org.apache.spark.sql.catalyst.plans.LeftAnti
 import org.apache.spark.sql.catalyst.plans.LeftSemi
 import org.apache.spark.sql.execution.SparkPlan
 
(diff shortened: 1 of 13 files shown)


src/main/scala/edu/berkeley/cs/rise/opaque/strategies.scala
```diff
@@ -76,7 +79,12 @@ object OpaqueOperators extends Strategy {
...
       val rightProj = EncryptedProjectExec(rightProjSchema, planLater(right))
       val unioned = EncryptedUnionExec(leftProj, rightProj)
-      val sorted = EncryptedSortExec(sortForJoin(leftKeysProj, tag, unioned.output), true, unioned)
+      // We partition based on the join keys only, so that rows from both the left and the right tables that match
+      // will colocate to the same partition
+      val partitionOrder = leftKeysProj.map(k => SortOrder(k, Ascending))
+      val partitioned = EncryptedRangePartitionExec(partitionOrder, unioned)
+      val sortOrder = sortForJoin(leftKeysProj, tag, partitioned.output)
+      val sorted = EncryptedSortExec(sortOrder, false, partitioned)
       val joined = EncryptedSortMergeJoinExec(
```
octaviansima, line 88: It looks like this case class no longer has a need for the `output` class attribute

octaviansima (approved):
LGTM, small comments

PR #155, opened 2021-02-10
Set up Github actions

octaviansima:
Migrate from Travis CI


octaviansima:
see https://github.com/mc2-project/opaque/pull/156

PR #157, opened 2021-02-13
Support for scalar subquery

wzheng:
This PR implements the scalar subquery expression, which is triggered whenever a subquery returns a scalar value. There were two main problems that needed to be solved.

First, support for matching the scalar subquery expression is necessary. Spark implements this by wrapping a `SparkPlan` within the expression and calls `executeCollect`. Then it constructs a literal with that value. However, this is problematic for us because that value should *not* be decrypted by the driver and serialized into an expression, since it's an intermediate value. 

Therefore, the second issue to be addressed here is supporting an *encrypted* literal. This is implemented in this PR by serializing an encrypted ciphertext into a base64 encoded string, and wrapping a `Decrypt` expression on top of it. This expression is then evaluated in the enclave and returns a literal. Note that, in order to test our implementation, we also implement a `Decrypt` expression in Scala. However, this should never be evaluated on the driver side and serialized into a plaintext literal. This is because `Decrypt` is designated as a `Nondeterministic` expression, and therefore will always evaluate on the workers.


src/test/scala/edu/berkeley/cs/rise/opaque/OpaqueOperatorTests.scala
```diff
@@ -879,6 +880,30 @@ trait OpaqueOperatorTests extends OpaqueTestsBase { self =>
...
   }
 
+  testAgainstSpark("encrypted literal") { securityLevel =>
+    val input = 10
+    val enc_str = Utils.encryptScalar(input, IntegerType)
+
+    val data = for (i <- 0 until 256) yield (i, abc(i), 1)
+    val words = makeDF(data, securityLevel, "id", "word", "count")
+    val df = words.filter($"id" < decrypt(lit(enc_str), IntegerType)).sort($"id")
+    df.collect()
```
octaviansima, line 890:
```suggestion
    df.collect
```
Just to keep consistent style

src/main/scala/edu/berkeley/cs/rise/opaque/expressions/Decrypt.scala
```diff
@@ -0,0 +1,52 @@
...
+    _FUNC_(child, outputDataType) - Decrypt the input evaluated expression, which should always be a string
+  """,
+  arguments = """
+    Arguments:
+      * child - an encrypted literal of string type
+      * outputDataType - the decrypted data type
+  """)
+/**
+ * 
+ */
```
octaviansima, line 33: Remove these empty comments or add a description?

src/main/scala/edu/berkeley/cs/rise/opaque/execution/operators.scala
```diff
@@ -134,14 +134,19 @@ trait OpaqueOperatorExec extends SparkPlan {
...
-   */
+    */
+
   override def doExecute(): RDD[InternalRow] = {
     sqlContext.sparkContext.emptyRDD
     // throw new UnsupportedOperationException("use executeBlocked")
   }
 
+  def collectEncrypted(): Array[Block] = {
+    executeBlocked().collect()
```
octaviansima, line 145:
```suggestion
    executeBlocked().collect
```

src/main/scala/edu/berkeley/cs/rise/opaque/Utils.scala
```diff
@@ -1112,6 +1168,37 @@ object Utils extends Logging {
...
+            case _ => throw new OpaqueException("Scalar subquery cannot match to AttributeReference")
+          }
+          // Need to deserialize the encrypted blocks to get the encrypted block
+          val blockList = child.asInstanceOf[OpaqueOperatorExec].collectEncrypted()
+          val encryptedBlocksList = blockList.map { block =>
+            val buf = ByteBuffer.wrap(block.bytes)
+            tuix.EncryptedBlocks.getRootAsEncryptedBlocks(buf)
+          }
+          val encryptedBlocks = encryptedBlocksList.find(_.blocksLength > 0).getOrElse(encryptedBlocksList(0))
+          println(s"encryptedBlocks = ${encryptedBlocks}")
```
octaviansima, line 1184: Remove `println`

src/main/scala/edu/berkeley/cs/rise/opaque/Utils.scala
```diff
@@ -822,6 +872,12 @@ object Utils extends Logging {
             tuix.ExprUnion.Literal,
             tuix.Literal.createLiteral(builder, valueOffset))
 
+        case (Decrypt(child, dataType), Seq(childOffset)) =>
```
octaviansima, line 875: Maybe include a comment stating that this should never be evaluated on the driver?

src/enclave/Enclave/util.h
```diff
@@ -41,4 +41,6 @@ int pow_2(int value);
 
 int secs_to_tm(long long t, struct tm *tm);
 
+std::string ciphertext_base64_decode(std::string const &encoded_string);
```
octaviansima, line 44:
```suggestion
std::string ciphertext_base64_decode(const std::string &encoded_string);
```
To keep style consistent

src/enclave/Enclave/ExpressionEvaluation.h
```diff
@@ -288,6 +288,49 @@ class FlatbuffersExpressionEvaluator {
...
+        std::vector<uint8_t> str_vec(
+            flatbuffers::VectorIterator<uint8_t, uint8_t>(str_field->value()->Data(),
+                                                          static_cast<uint32_t>(0)),
+            flatbuffers::VectorIterator<uint8_t, uint8_t>(str_field->value()->Data(),
+                                                          static_cast<uint32_t>(str_field->length())));
+
+        std::string ciphertext(str_vec.begin(), str_vec.end());
+        std::string ciphertext_decoded = ciphertext_base64_decode(ciphertext);
+
+        uint8_t *plaintext = new uint8_t[ciphertext_decoded.size()];
```
octaviansima, line 316:
```suggestion
        uint8_t *plaintext = new uint8_t[dec_size(ciphertext_decoded.size())];
```

octaviansima (approved):
LGTM! Minor comments

PR #159, opened 2021-02-18
Broadcast Nested Loop Join - Left Anti and Left Semi

octaviansima:
This PR is the first of two parts towards making TPC-H 16 work: the other will be implementing `is_distinct` for aggregate operations.

`BroadcastNestedLoopJoin` is Spark's "catch all" for non-equi joins. It works by first picking a side to broadcast, then iterating through every possible row combination and checking the non-equi condition against the pair.

### DIFF
--- a/src/enclave/Enclave/BroadcastNestedLoopJoin.cpp
+++ b/src/enclave/Enclave/BroadcastNestedLoopJoin.cpp
@@ -1,0 +1,54 @@
+#include "BroadcastNestedLoopJoin.h"
+
+#include "ExpressionEvaluation.h"
+#include "FlatbuffersReaders.h"
+#include "FlatbuffersWriters.h"
+#include "common.h"
+
+/** C++ implementation of a broadcast nested loop join.
+ * Assumes outer_rows is streamed and inner_rows is broadcast.
+ * DOES NOT rely on rows to be tagged primary or secondary, and that
+ * assumption will break the implementation.
+ */
+void broadcast_nested_loop_join(
+  uint8_t *join_expr, size_t join_expr_length,
+  uint8_t *outer_rows, size_t outer_rows_length,
+  uint8_t *inner_rows, size_t inner_rows_length,
+  uint8_t **output_rows, size_t *output_rows_length) {
+
+  FlatbuffersJoinExprEvaluator join_expr_eval(join_expr, join_expr_length);
+  const tuix::JoinType join_type = join_expr_eval.get_join_type();
+
+  RowReader outer_r(BufferRefView<tuix::EncryptedBlocks>(outer_rows, outer_rows_length));
+  RowWriter w;
+
+  while (outer_r.has_next()) {
+    const tuix::Row *outer = outer_r.next();
+    bool o_i_match = false;
+
+    RowReader inner_r(BufferRefView<tuix::EncryptedBlocks>(inner_rows, inner_rows_length));
+    const tuix::Row *inner;
+    while (inner_r.has_next()) {
+      inner = inner_r.next();
+      o_i_match |= join_expr_eval.eval_condition(outer, inner);
+    }
+
+    switch(join_type) {
+      case tuix::JoinType_LeftAnti:
+        if (!o_i_match) {
+          w.append(outer);
+        }
+        break;
+      case tuix::JoinType_LeftSemi:
+        if (o_i_match) {
+          w.append(outer);
+        }
+        break;
+      default:
+        throw std::runtime_error(
+          std::string("Join type not supported: ")
+          + std::string(to_string(join_type)));
+    }
+  }
+  w.output_buffer(output_rows, output_rows_length);
+}
--- a/src/enclave/Enclave/ExpressionEvaluation.h
+++ b/src/enclave/Enclave/ExpressionEvaluation.h
@@ -1725,59 +1725,103 @@ public:
     }
 
     const tuix::JoinExpr* join_expr = flatbuffers::GetRoot<tuix::JoinExpr>(buf);
-    join_type = join_expr->join_type();
 
-    if (join_expr->left_keys()->size() != join_expr->right_keys()->size()) {
-      throw std::runtime_error("Mismatched join key lengths");
+    join_type = join_expr->join_type();
+    if (join_expr->condition() != NULL) {
+      condition_eval = std::unique_ptr<FlatbuffersExpressionEvaluator>(
+          new FlatbuffersExpressionEvaluator(join_expr->condition()));
     }
-    for (auto key_it = join_expr->left_keys()->begin();
-         key_it != join_expr->left_keys()->end(); ++key_it) {
-      left_key_evaluators.emplace_back(
-        std::unique_ptr<FlatbuffersExpressionEvaluator>(
-          new FlatbuffersExpressionEvaluator(*key_it)));
-    }
-    for (auto key_it = join_expr->right_keys()->begin();
-         key_it != join_expr->right_keys()->end(); ++key_it) {
-      right_key_evaluators.emplace_back(
-        std::unique_ptr<FlatbuffersExpressionEvaluator>(
-          new FlatbuffersExpressionEvaluator(*key_it)));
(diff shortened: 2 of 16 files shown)


src/main/scala/edu/berkeley/cs/rise/opaque/strategies.scala
```diff
@@ -183,17 +208,29 @@ object OpaqueOperators extends Strategy {
...
+    joinType match {
+      case LeftSemiOrAnti(joinType) => BuildRight
+      case _ => BuildLeft
+    }
+  }
+
+  // Everything below is a private method in SparkStrategies.scala
+  private def getSmallerSide(left: LogicalPlan, right: LogicalPlan): BuildSide = {
+    if (right.stats.sizeInBytes <= left.stats.sizeInBytes) BuildRight else BuildLeft
+  }
```
wzheng, line 235: The original method seems to be public, from `joins.scala`?

src/main/scala/edu/berkeley/cs/rise/opaque/strategies.scala
```diff
@@ -105,6 +111,25 @@ object OpaqueOperators extends Strategy {
...
 
+    // Used to match non-equi joins
+    case Join(left, right, joinType, condition, hint) if isEncrypted(left) && isEncrypted(right) =>
+      // For perf reasons, `BroadcastNestedLoopJoinExec` prefers to broadcast left side if
+      // it's a right join, and broadcast right side if it's a left join.
+      // TODO: revisit it. If left side is much smaller than the right side, it may be better
+      // to broadcast the left side even if it's a left join.
+      val desiredBuildSide = if (joinType.isInstanceOf[InnerLike] || joinType == FullOuter)
+          getSmallerSide(left, right) else
+          getBroadcastSideBNLJ(joinType)
```
wzheng, line 122: Can you use the Spark SQL function `canBuildBroadcastLeft`?

src/main/scala/edu/berkeley/cs/rise/opaque/strategies.scala
```diff
@@ -105,6 +111,25 @@ object OpaqueOperators extends Strategy {
 
       filtered :: Nil
 
+    // Used to match non-equi joins
+    case Join(left, right, joinType, condition, hint) if isEncrypted(left) && isEncrypted(right) =>
+      // For perf reasons, `BroadcastNestedLoopJoinExec` prefers to broadcast left side if
+      // it's a right join, and broadcast right side if it's a left join.
+      // TODO: revisit it. If left side is much smaller than the right side, it may be better
+      // to broadcast the left side even if it's a left join.
```
wzheng, line 119: Nit: this seems to be a comment copied from Spark, maybe can just summarize their comment?

src/flatbuffers/operators.fbs
```diff
@@ -54,10 +54,11 @@ enum JoinType : ubyte {
...
-    // parallel arrays of key expressions and the join outputs pairs of rows
-    // where each expression from the left is equal to the matching expression
-    // from the right.
+    // In the case of equi joins, we store parallel arrays of key expressions and have the join output
+    // pairs of rows  where each expression from the left is equal to the matching expression from the right.
     left_keys:[Expr];
     right_keys:[Expr];
+    // In the case of non-equi joins, we pass in a condition
+    // as an expression and evaluate that on each pair of rows.
+    condition:Expr;
```
wzheng, line 63: Equijoin also supports condition, which we currently handle using an extra filter operation. We might want to merge that into the join code at some point instead of using an extra operator. Doesn't have to be in this PR, but might be good to put a TODO comment here?

wzheng (commented):
Partial review.

src/main/scala/edu/berkeley/cs/rise/opaque/execution/operators.scala
```diff
@@ -308,6 +307,73 @@ case class EncryptedSortMergeJoinExec(
...
+    }
+  }
+
+  override def executeBlocked(): RDD[Block] = {
+    val joinExprSer = Utils.serializeJoinExpression(
+        joinType, None, None, left.output, right.output, condition)
+
+    val leftRDD = left.asInstanceOf[OpaqueOperatorExec].executeBlocked()
+    val rightRDD = right.asInstanceOf[OpaqueOperatorExec].executeBlocked()
+
```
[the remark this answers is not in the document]
octaviansima: This is what Spark does but I found the existing implementation much easier and shorter. The issue with (streamed, broadcast) is that I would have to convert LeftAnti to RightAnti, etc. for the C++ code otherwise the result with contain the wrong columns.

src/main/scala/edu/berkeley/cs/rise/opaque/strategies.scala
octaviansima: Not in Spark 3.0.1

src/main/scala/edu/berkeley/cs/rise/opaque/strategies.scala
octaviansima: Not in Spark 3.0.1

src/enclave/Enclave/ExpressionEvaluation.h
```diff
@@ -1725,60 +1725,103 @@ class FlatbuffersJoinExprEvaluator {
...
+    return is_primary(row1) ? row1 : row2;
+  }
+
   /** Return true if the two rows are from the same join group. */
-  bool is_same_group(const tuix::Row *row1, const tuix::Row *row2) {
+  bool eval_condition(const tuix::Row *row1, const tuix::Row *row2) {
+    builder.Clear();
+    bool row1_equals_row2;
+
+    /* Check equality for equi joins */
```
wzheng, line 1782: This should only be called if `is_equi_join` is true?

src/enclave/Enclave/BroadcastNestedLoopJoin.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+    RowReader inner_r(BufferRefView<tuix::EncryptedBlocks>(inner_rows, inner_rows_length));
+    const tuix::Row *inner;
+    while (inner_r.has_next()) {
+      inner = inner_r.next();
+      o_i_match |= join_expr_eval.eval_condition(outer, inner);
+    }
+
+    switch(join_type) {
+      case tuix::JoinType_LeftAnti:
```
wzheng, line 40: Can you support general left existence join here, with LeftSemi as well?

src/main/scala/edu/berkeley/cs/rise/opaque/execution/operators.scala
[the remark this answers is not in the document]
octaviansima: Correct: left side always corresponds to `outer_rows` and right side always corresponds to `inner_rows` in C++. I think the strongest case for keeping the code as is comes from strategies.scala: Under the current scheme, if we find a significant performance boost for, as an example, building left when the join is `LeftSemi` (could be due to left side being much smaller), then there will be no required changes to operators.scala or the C++ code. If we decide to model after the Spark code, then switching `outer_rows` and `inner_rows` will mean implementing `RightSemi` in C++ even though it is not officially supported by Scala.

src/enclave/Enclave/ExpressionEvaluation.h
octaviansima: Yes, but the code never actually goes into the for loop because `row1_evaluators` is empty.

src/enclave/Enclave/BroadcastNestedLoopJoin.cpp
octaviansima: LeftSemi done, but it looks like left existence is not supported? https://github.com/apache/spark/blob/429f8af9b683935151c2379bc80b27162cd1c8bf/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/joinTypes.scala#L90

wzheng (approved):
Looks great, thanks!

PR #160, opened 2021-02-19
Float expressions

wzheng:
This PR adds float normalization expressions [implemented in Spark](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/NormalizeFloatingNumbers.scala#L170). TPC-H query 2 also passes.

### DIFF
--- a/src/main/scala/edu/berkeley/cs/rise/opaque/Utils.scala
+++ b/src/main/scala/edu/berkeley/cs/rise/opaque/Utils.scala
@@ -59,6 +59,7 @@ import org.apache.spark.sql.catalyst.expressions.If
 import org.apache.spark.sql.catalyst.expressions.In
 import org.apache.spark.sql.catalyst.expressions.IsNotNull
 import org.apache.spark.sql.catalyst.expressions.IsNull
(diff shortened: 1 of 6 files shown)


src/main/scala/edu/berkeley/cs/rise/opaque/Utils.scala
```diff
@@ -1199,6 +1210,7 @@ object Utils extends Logging {
             )
           }
 
+
```
octaviansima, line 1213:
```suggestion

```

src/test/scala/edu/berkeley/cs/rise/opaque/OpaqueOperatorTests.scala
```diff
@@ -344,6 +344,15 @@ trait OpaqueOperatorTests extends OpaqueTestsBase { self =>
     df.collect
   }
 
+  testAgainstSpark("join on floats") { securityLevel =>
```
octaviansima, line 347: I would try to increase the coverage by adding some null values and potentially another test for doubles.

octaviansima (approved):
LGTM!

PR #162, opened 2021-02-23
Left/Right Outer support for equi and non-equi joins

octaviansima:
TPC-H 13 passes.

This PR implements left and right outer for both physical operators of join that are currently supported: `NonObliviousSortMergeJoin` and `BroadcastNestedLoopJoin`. It adds tests for both, but TPC-H 13 only requires the equi join left outer implementation.


src/enclave/Enclave/NonObliviousSortMergeJoin.cpp
```diff
@@ -128,20 +156,27 @@ void non_oblivious_sort_merge_join(
...
-    write_output_rows(primary_matched_rows, w);
-  } else if (join_type == tuix::JoinType_LeftAnti) {
-    write_output_rows(primary_unmatched_rows, w);
+  switch (join_type) {
+    case tuix::JoinType_LeftSemi:
+      write_output_rows(primary_matched_rows, w, join_type);
+      break;
+    case tuix::JoinType_LeftAnti:
+      write_output_rows(primary_unmatched_rows, w, join_type);
+      write_output_rows(previous_primary_unmatched_rows, w, join_type);
```
wzheng, line 171: If the input data is very big, then won't `previous_primary_unmatched_rows` potentially have to buffer a lot of data, and it will have to write out to `w` again? Can you write out to `w` directly earlier?

octaviansima: The issue with doing it earlier is that `last_foreign_row.get()` can be null. This happens when a new primary group is encountered without any foreign row being encoounted first at all.

wzheng: For outer joins, can you put the dummy row first instead of last?

octaviansima: that worked, thanks

wzheng (approved):
Thanks!

PR #163, opened 2021-02-24
Distinct aggregation support

octaviansima:
TPC-H 16 passes.

This is part 2 of making TPC-H 16 work.

Currently, our code does not support aggregate expressions that have `isDistinct` set to true. This PR intends to add support for all valid aggregate expressions (`Count` and `Sum`, but query 16 only requires `Count`).

This PR differs from https://github.com/mc2-project/opaque/pull/161 in one major way: there are no changes to the C++ code. The implementation for `isDistinct` is done completely in Scala by chaining together 4 stages of aggregations. This works by performing intermediate aggregations that result in the distinct operation having purely unique arguments. . For reference, see [the Spark implementation](https://github.com/apache/spark/blob/b17754a8cbd2593eb2b1952e95a7eeb0f8e09cdb/sql/core/src/main/scala/org/apache/spark/sql/execution/aggregate/AggUtils.scala#L134.).

### DIFF
--- a/src/main/scala/edu/berkeley/cs/rise/opaque/execution/operators.scala
+++ b/src/main/scala/edu/berkeley/cs/rise/opaque/execution/operators.scala
@@ -233,69 +233,29 @@ case class EncryptedFilterExec(condition: Expression, child: SparkPlan)
 
 case class EncryptedAggregateExec(
   groupingExpressions: Seq[NamedExpression],
-  aggExpressions: Seq[AggregateExpression],
-  // Specify mode if ALL aggExpressions should be cast to the provided AggregateMode
-  mode: Option[AggregateMode],
+  aggregateExpressions: Seq[AggregateExpression],
   child: SparkPlan)
     extends UnaryExecNode with OpaqueOperatorExec {
 
   override def producedAttributes: AttributeSet =
-    AttributeSet(aggExpressions) -- AttributeSet(groupingExpressions)
+    AttributeSet(aggregateExpressions) -- AttributeSet(groupingExpressions)
 
-  override def output: Seq[Attribute] = mode match {
-    case Some(Partial) =>
-      groupingExpressions.map(_.toAttribute) ++
-        aggExpressions.map(_.copy(mode = Partial)).flatMap(_.aggregateFunction.inputAggBufferAttributes)
-    case Some(PartialMerge) =>
-      groupingExpressions.map(_.toAttribute) ++
-        aggExpressions.map(_.copy(mode = PartialMerge)).flatMap(_.aggregateFunction.inputAggBufferAttributes)
-    case Some(Final) =>
-      groupingExpressions.map(_.toAttribute) ++ aggExpressions.map(_.resultAttribute)
-    case Some(Complete) =>
-      groupingExpressions.map(_.toAttribute) ++ aggExpressions.map(_.resultAttribute)
-    case None =>
-      groupingExpressions.map(_.toAttribute) ++ aggExpressions.flatMap(expr => {
-        expr.mode match {
-          case Partial | PartialMerge =>
-            expr.aggregateFunction.inputAggBufferAttributes
-          case _ =>
-            Seq(expr.resultAttribute)
-        }
-      })
-  }
+  override def output: Seq[Attribute] = groupingExpressions.map(_.toAttribute) ++
+    aggregateExpressions.flatMap(expr => {
+      expr.mode match {
+        case Partial | PartialMerge =>
+          expr.aggregateFunction.inputAggBufferAttributes
+        case _ =>
+          Seq(expr.resultAttribute)
+      }
+    })
 
   override def executeBlocked(): RDD[Block] = {
 
-    val (groupingExprs, aggExprs) = mode match {
-      case Some(Partial) => {
-        val partialAggExpressions = aggExpressions.map(_.copy(mode = Partial))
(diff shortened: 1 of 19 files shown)


src/main/scala/edu/berkeley/cs/rise/opaque/strategies.scala
```diff
@@ -136,25 +136,84 @@ object OpaqueOperators extends Strategy {
...
+            // Indistinct functions operate on aggregation buffers since partial aggregation was already called,
+            // but distinct functions operate on the original input to the aggregation.
+            EncryptedAggregateExec(groupingExpressions,
+              functionsWithoutDistinct.map(_.copy(mode = PartialMerge)) ++
+                functionsWithDistinct.map(_.copy(mode = Partial)), None, partialMergeAggregate)
+          }
+
+          // 4. Create an Aggregate operator for the final aggregation.
+          val finalAggregate = {
+            val sorted = EncryptedSortExec(groupingExpressions.map(e => SortOrder(e, Ascending)),
```
octaviansima, line 206: Unfortunately, I could not find a way to get rid of this global sort. Without it, the result is off by the following margin: `Set([B,4,11], [C,4,10], [A,4,11]) did not equal Set([C,1,2], [B,3,8], [B,1,3], [C,3,8], [A,4,11])`

src/main/scala/edu/berkeley/cs/rise/opaque/strategies.scala
```diff
@@ -136,25 +136,84 @@ object OpaqueOperators extends Strategy {
...
+              true, partialDistinctAggregate)
```
octaviansima, line 207: Unfortunately, I could not find a way to get rid of this global sort. Without it, the result is off by the following margin: `Set([B,4,11], [C,4,10], [A,4,11]) did not equal Set([C,1,2], [B,3,8], [B,1,3], [C,3,8], [A,4,11])`

octaviansima, line 207: Unfortunately, I could not find a way to get rid of this global sort. Without it, the result is off by the following margin: `Set([B,4,11], [C,4,10], [A,4,11]) did not equal Set([C,1,2], [B,3,8], [B,1,3], [C,3,8], [A,4,11])`

src/main/scala/edu/berkeley/cs/rise/opaque/execution/operators.scala
```diff
@@ -234,42 +234,73 @@ case class EncryptedFilterExec(condition: Expression, child: SparkPlan)
 case class EncryptedAggregateExec(
   groupingExpressions: Seq[NamedExpression],
   aggExpressions: Seq[AggregateExpression],
-  mode: AggregateMode,
+  // Specify mode if ALL aggExpressions should be cast to the provided AggregateMode
+  mode: Option[AggregateMode],
```
wzheng, line 238: Seems like mode isn't really needed if this information is embedded in `aggExpressions`?

src/test/scala/edu/berkeley/cs/rise/opaque/OpaqueOperatorTests.scala
```diff
@@ -461,6 +461,20 @@ trait OpaqueOperatorTests extends OpaqueTestsBase { self =>
       .collect.sortBy { case Row(category: String, _) => category }
   }
 
+  testAgainstSpark("aggregate count distinct and indistinct") { securityLevel =>
```
wzheng, line 464: Would be good to test for NULL values as well.

src/main/scala/edu/berkeley/cs/rise/opaque/strategies.scala
wzheng: I think it's because in the first global sort (in the second step), the sort is executed on `combinedGroupingExpressions`? This does not guarantee that the rows belonging to the same group under `groupingExpressions` only will be colocated to the same partition. You can solve it by first doing an explicit range partitioning on `groupingExpressions`, followed by a per partition sort on `combinedGroupingExpressions`. A similar strategy is currently used in join: https://github.com/mc2-project/opaque/blob/master/src/main/scala/edu/berkeley/cs/rise/opaque/strategies.scala#L92.

wzheng (approved):
Looks great, thanks!

src/main/scala/edu/berkeley/cs/rise/opaque/strategies.scala
```diff
@@ -132,25 +132,90 @@ object OpaqueOperators extends Strategy {
...
-      } else {
-        // Grouping aggregation
-        EncryptedProjectExec(resultExpressions,
-          EncryptedAggregateExec(groupingExpressions, aggregateExpressions, Final,
-            EncryptedSortExec(groupingExpressions.map(_.toAttribute).map(e => SortOrder(e, Ascending)), true,
-              EncryptedAggregateExec(groupingExpressions, aggregateExpressions, Partial,
-                EncryptedSortExec(groupingExpressions.map(e => SortOrder(e, Ascending)), false, planLater(child)))))) :: Nil
+      val (functionsWithDistinct, functionsWithoutDistinct) = aggregateExpressions.partition(_.isDistinct)
+
+      functionsWithDistinct.size match {
```
Oliver001, line 137: why not  functionsWithDistinct.map(_.aggregateFunction.children.toSet).distinct.length?

can we support this query:select count(distinct age) ,sum(distinct age) from table_a?

PR #164, opened 2021-02-25
Move join condition handling for equi-joins into enclave code

wzheng:
This PR moves the condition processing code for equi-joins into the enclave side. This allows us to handle the condition statement with one fewer operator. It also solves a bug in LeftSemi/Anti joins caused by the condition being executed as a filter operation on the left table, when it should be executed on the joined row.

This PR also enables TPC-H 21.


octaviansima (changes requested):
Small changes to improve readability of the C++ code.

octaviansima (approved):
LGTM! Thanks

PR #165, opened 2021-02-27
Benchmark code tests for correctness as well

octaviansima:
This PR enables the benchmarking code to check for correctness by building it alongside tests. In doing so, there needed to be a new sbt operation in the form of `test:assembly` that creates a jar containing classes from both src and test in order for Spark to be able to run the benchmarking code. There is also a new section to the readme explaining how to run benchmarking from the command-line.


src/test/scala/edu/berkeley/cs/rise/opaque/benchmark/Benchmark.scala
```diff
@@ -44,12 +44,12 @@ object Benchmark {
 
   val spark = SparkSession.builder()
       .appName("Benchmark")
+      .master("local[4]")
```
wzheng, line 47: Can this be used for the distributed benchmark?

octaviansima:
Contains a couple of files that have no functional changes but were linted. They weren't picked up in https://github.com/mc2-project/opaque/pull/166 because I merged a PR in right after without committing anything first.

wzheng (approved):
Thanks!

PR #167, opened 2021-03-08
Upgrade Spark to 3.1.1

octaviansima:


src/main/scala/edu/berkeley/cs/rise/opaque/strategies.scala
```diff
@@ -381,18 +381,8 @@ object OpaqueOperators extends Strategy {
     (Seq(tag) ++ input, tag.toAttribute)
   }
 
-  private def getBroadcastSideBNLJ(joinType: JoinType): BuildSide = joinType match {
-    case LeftExistence(_) | LeftOuter => BuildRight
-    case _ => BuildLeft
-  }
-
```
wzheng, line 388: I wonder if we actually want to keep this definition, in case Spark SQL changes the definition and no longer satisfies our constraints?

wzheng (approved):
Just one comment!

PR #170, opened 2021-03-09
Robust attestation

wzheng:
This PR contains a more fault tolerant way to do attestation. The remote attestation code runs in a loop in a separate thread in order to constantly check for unattested machines. This is done through the use of an accumulator variable -- every time a new enclave is started, the accumulator variable's state is changed to indicate that attestation needs to be re-run.

One caveat is that this attestation design relies on Spark's fault tolerance design. If a task attempts to run on an executor with an unattested enclave, the code will throw an exception. Therefore, the `spark.task.maxFailures` must be set to a number greater than 1 so that Spark will rerun the task. Currently this configuration variable is set to 10 for the tests.


src/main/scala/edu/berkeley/cs/rise/opaque/Utils.scala
```diff
@@ -293,18 +279,94 @@ object Utils extends Logging {
...
+        val enclave = new SGXEnclave()
+        (enclave, eid)
+      }
+    }
+
+  }  
+
+  def generateReport(): (Long, Option[Array[Byte]]) = {
+    this.synchronized {
+      if (!attested) {
```
octaviansima, line 344: The solution I came up with involved adding:
```.scala
if (eid == 0) {
 Thread.sleep(200)
  throw new OpaqueException("Enclave has not been started")
}
```
and just letting the fault tolerance handle it. With this, there should be no need to have the `wait` function anymore.

octaviansima (changes requested):
I commented the fix for un-started enclaves that relies on the fault tolerance. Since the code will throw that exception at most `numExecutors` times, I think it would be fine. It also allows us to get rid of the `wait` function completely. We can discuss whether or not this solution is the one that should be accepted.

octaviansima:
Found another alternative solution that has a correct `wait`:

```scala
  def wait(sc: SparkContext): Unit = {
    if (!sc.isLocal) {
      numExecutors = sc.getConf.getInt("spark.executor.instances", -1)
      val rdd = sc.parallelize(Seq.fill(numExecutors) { () }, numExecutors)
      while (
        (rdd
          .mapPartitions { (_) =>
            val id = SparkEnv.get.executorId
            Iterator(id)
          }
          .collect
          .toSet)
          .size < numExecutors
      ) {}
    }
    logInfo(s"All executors have started, numExecutors is ${numExecutors}")
  }
```

src/main/scala/edu/berkeley/cs/rise/opaque/Utils.scala
wzheng: I changed this to not fail, and instead generate a report only when the enclave has been started.

src/main/scala/edu/berkeley/cs/rise/opaque/Utils.scala
```diff
@@ -293,18 +279,94 @@ object Utils extends Logging {
...
   def initSQLContext(sqlContext: SQLContext): Unit = {
     sqlContext.experimental.extraOptimizations =
       (Seq(EncryptLocalRelation, ConvertToOpaqueOperators) ++
         sqlContext.experimental.extraOptimizations)
     sqlContext.experimental.extraStrategies = (Seq(OpaqueOperators) ++
       sqlContext.experimental.extraStrategies)
-    RA.initRA(sqlContext.sparkContext)
+
+    val sc = sqlContext.sparkContext
+    if (!acc_registered) {
```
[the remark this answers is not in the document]
wzheng: This is to prevent the Utils object from reused across multiple tests, and causing an error because the accumulator variables are re-registered. I added a comment about this.

src/main/scala/edu/berkeley/cs/rise/opaque/RA.scala
```diff
@@ -69,5 +74,36 @@ object RA extends Logging {
...
+      Iterator(eid)
+    }.collect
+
+    if (Utils.numUnattested.value != Utils.numAttested.value) {
+      logInfo(
+        s"RA.run: ${Utils.numUnattested.value} unattested, ${Utils.numAttested.value} attested"
+      )
+      initRA(sc)
+    }
+    Thread.sleep(100)
```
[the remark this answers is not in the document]
wzheng: I changed some code here. The loop is still in RA.scala, but the thread is also being started in there.

.hooks/pre-commit
```diff
@@ -39,4 +39,3 @@ then
 fi
 
 echo "Pre-commit hook finished."
-
```
octaviansima, line 42: nit: revert

src/main/scala/edu/berkeley/cs/rise/opaque/RA.scala
```diff
@@ -26,48 +26,104 @@ import edu.berkeley.cs.rise.opaque.execution.SP
...
     // Runs on driver
-    val msg2s = msg1s.map { case (eid, msg1) => (eid, sp.ProcessEnclaveReport(msg1)) }
+    val msg2s = msg1s.collect { case (eid, Some(msg1: Array[Byte])) =>
+      (eid, sp.ProcessEnclaveReport(msg1))
+    }
+
+    logInfo("Driver processed msg2s")
 
     // Runs on executors
     val attestationResults = rdd
```
octaviansima, line 65: This variable is unused. Either remove it or add
```scala
    for ((_, ret) <- attestationResults) {
      if (!ret)
        throw new OpaqueException("Attestation failed")
    }
```

src/main/scala/edu/berkeley/cs/rise/opaque/Utils.scala
```diff
@@ -293,18 +279,91 @@ object Utils extends Logging {
...
+        val msg2 = msg2s(eid)
+        enclave.FinishAttestation(eid, msg2)
+        numAttested.add(1)
+        attested = true
+      }
+      (eid, attested)
+    }
+  }
+
+  def cleanup() {
```
octaviansima, line 365: I would recommend include `spark.stop()` here so you don't have to always couple those two functions together.

octaviansima (approved):
Looks great, thanks!

PR #171, opened 2021-03-09
Test refactor: Query operations

octaviansima:
Part 1 of 2 of the new testing infrastructure. Part 2 will include all Opaque functionality that cannot be tested against Spark (EncryptedSource, etc.)

This PR includes new tests for all the existing query operations that are supported, as well as those that are not yet supported (ignored). It also refactors much of the existing infrastructure, and defines new base classes that simplify creating tests for the plain DataFrame API as well as SQL queries. All queries are taken directly from Spark's own testing.


src/test/scala/edu/berkeley/cs/rise/opaque/AggregationSuite.scala
```diff
@@ -0,0 +1,492 @@
...
+
+    checkAnswer() { sl =>
+      loadAggData(sl)
+      spark.sql("""
+          |SELECT avg(value) FROM agg1
+        """.stripMargin)
+    }
+  }
+
+  ignore("first_value and last_value") {
```
wzheng, line 233: Will we have support for this in the future?

src/test/scala/edu/berkeley/cs/rise/opaque/JoinSuite.scala
```diff
@@ -0,0 +1,357 @@
...
+
+  test("inner join, no matches") {
+    checkAnswer() { sl =>
+      val x = testData2(sl).where($"a" === 1).as("x")
+      val y = testData2(sl).where($"a" === 2).as("y")
+      x.join(y).where($"x.a" === $"y.a")
+    }
+  }
+
+  ignore("big inner join, 4 matches per row") {
```
wzheng, line 88: Why is this ignored?

src/test/scala/edu/berkeley/cs/rise/opaque/QEDSuite.scala
```diff
@@ -18,22 +18,14 @@
 package edu.berkeley.cs.rise.opaque
```
wzheng, line 18: Maybe we should rename `QEDSuite` -- it was an old name for Opaque :P

src/test/scala/edu/berkeley/cs/rise/opaque/AggregationSuite.scala
octaviansima: I would think so? Seems like a function that is supported by Spark SQL https://spark.apache.org/docs/3.1.1/api/sql/index.html#first_value

src/test/scala/edu/berkeley/cs/rise/opaque/JoinSuite.scala
octaviansima: Casting error :( 
`java.lang.ClassCastException: org.apache.spark.sql.execution.UnionExec cannot be cast to edu.berkeley.cs.rise.opaque.execution.OpaqueOperatorExec`
Should be a supported query, but looks we need to fix some bugs.

PR #182, opened 2021-03-17
Error handling for missing functionalities

wzheng:
This PR introduces better error handling for missing operators and expressions. For operators, the planning stage in strategies will throw an error if there is an unmatched logical operator. For expressions, then existing code will already fail early because `executeBlocked()` will run immediately for every operator and return a lazily evaluated RDD. Therefore, the expressions will also be serialized before the RDD is eagerly evaluated.


octaviansima (approved):
Thanks!

One quick suggestion: "is not supported in Opaque" ->"is not _currently_ supported in Opaque" to make it clear that we do intend to support some of these operators down the road.

PR #184, opened 2021-03-18
Add remaining tests

octaviansima:
Adds all relevant query tests from https://github.com/apache/spark/blob/master/sql/core/src/test/scala/org/apache/spark/sql/SubquerySuite.scala + Opaque specific tests and other miscellaneous operations.

We also add a new file, SharedSparkSessions.scala, that contains definitions of Spark sessions shared between test suites. Right now, each session is started and stopped at the beginning and end of each suite because of `OpaqueSuiteBase`. As the number of our suites grow, we should consider starting/stopping a session only once, for all test suites. This is because it takes a while for `local-cluster` to spin up a session; it would increase our performance. This can't be done at the moment because the cluster throws errors unless `Utils.cleanup(spark)` is called -- we need to figure out how to run this at the very end of all suites.


src/test/scala/edu/berkeley/cs/rise/opaque/MiscSuite.scala
```diff
@@ -0,0 +1,86 @@
...
+
+import org.apache.spark.sql.functions._
+
+import edu.berkeley.cs.rise.opaque.benchmark._
+import edu.berkeley.cs.rise.opaque.expressions.DotProduct.dot
+import edu.berkeley.cs.rise.opaque.expressions.VectorMultiply.vectormultiply
+import edu.berkeley.cs.rise.opaque.expressions.VectorSum
+
+/* Contains non-query related Spark tests */
+trait MiscSuite extends OpaqueSuiteBase {
```
wzheng, line 28: Maybe could call this the `OpaqueUDFSuite`?

src/test/scala/edu/berkeley/cs/rise/opaque/OpaqueSpecificSuite.scala
```diff
@@ -0,0 +1,167 @@
...
+      )
+    )
+    // Trigger an Opaque exception by attempting an unsupported cast: CalendarIntervalType to StringType
+    val e = intercept[SparkException] {
+      withLoggingOff {
+        df.select($"CalendarIntervalType".cast(StringType)).collect
+      }
+    }
+    assert(e.getCause.isInstanceOf[OpaqueException])
+  }
```
wzheng, line 166: I don't see the encrypted literal test, can you add it back?

wzheng (approved):
A couple of comments, thanks!

PR #186, opened 2021-03-23
Add support for PySpark

octaviansima:
Adds a Python file that has a wrapper for `Utils.initSQLContext` and creates a DataFrame attribute `encrypted`. Also adds relevant documentation to usage.rst.


docs/src/usage/usage.rst
```diff
@@ -18,79 +18,134 @@ Running Opaque SQL
...
-   3. Inside the Spark shell, import Opaque's DataFrame methods and install Opaque's query planner rules:
+                     import edu.berkeley.cs.rise.opaque.implicits._
+                     edu.berkeley.cs.rise.opaque.Utils.initSQLContext(spark.sqlContext)
 
-      .. code-block:: scala
+   Python:
+
+   .. code-block:: python
+
+                  from opaque_wrappers import *
```
wzheng, line 58: Maybe we can call it `opaque_sql` instead of `opaque_wrappers`?

wzheng (approved):
Looks great, thanks!

PR #188, opened 2021-03-24
Key gen

eric-feng-2011:
Adds local attestation between worker enclaves (coordinated by the driver) for deriving shared enclave key.
Removes keys from untrusted memory (service provider, driver, executor driver) so that all encryption/decryption happens in enclaves.


chester-leung:
can you please add a description of the workflow? i.e. a sequence of steps that are taken during local attestation, and the functions that are called? the PR is a bit large, and this will help with understanding all the code that you've added. thanks!

eric-feng-2011:
I wrote up a quick document about the steps and functions calls. I linked it in the original description, but here it is again for convenience:  https://drive.google.com/file/d/1EtHdv1WbyuTk4r6OGGWCNuufK6MC6SnA/view?usp=sharing

PR #197, opened 2021-04-05
IntpHandler: capture stdout and return to the client

octaviansima:
Have the client receive both the result returned by computation as well as any print statements (or other outputs to `Console.out`)

### DIFF
--- a/src/main/scala/edu/berkeley/cs/rise/opaque/rpc/IntpHandler.scala
+++ b/src/main/scala/edu/berkeley/cs/rise/opaque/rpc/IntpHandler.scala
@@ -99,7 +99,7 @@ object IntpHandler {
       val ret = out.getBuffer.toString
       (ret, res)
     }
-    (ret + "\n\n stdout: \n\n" + captured.toString, res)
+    (captured.toString + "\n----------------------------\n" + ret, res)
   }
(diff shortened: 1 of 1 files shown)


src/main/scala/edu/berkeley/cs/rise/opaque/rpc/IntpHandler.scala
```diff
@@ -92,10 +92,14 @@ object IntpHandler {
...
-    val output = out.getBuffer.toString
-    (output, res)
+    val captured = new ByteArrayOutputStream()
+    val (ret, res) = Console.withOut(captured) {
+      out.getBuffer.setLength(0)
+      val res = intp.interpret(input)
+      val ret = out.getBuffer.toString
+      (ret, res)
+    }
+    (captured.toString + "\n----------------------------\n" + ret, res)
```
wzheng, line 102: Can you add a comment on what `captured`, `ret`, and `res` mean?

wzheng (approved):

octaviansima:
I actually found a cleaner way to do this that routes all output to stdout automatically, so the only output stream needed is `captured`

wzheng (approved):

PR #198, opened 2021-04-05
Better assembly build

octaviansima:
This PR enables running the listener with `build/sbt assembly` instead of `build/sbt test:assembly` (basically submitting a jar to Spark that does not contain unnecessary test files). It also renames the shade rule to better reflect what we are doing (changing references to Guava classes for non-Spark dependencies to use a newer version). We actually do not have to selectively pick the libraries this should be done for, since Spark is marked as "Provided" automatically by the `sbt-spark` plugin, meaning it is not included in the fat jar created by `build/sbt assembly`.

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -4,6 +4,6 @@ addSbtPlugin("com.github.alonsodomin" % "sbt-spark" % "0.6.0")
 
 addSbtPlugin("ch.jodersky" % "sbt-jni" % "1.4.1")
 
-addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.15.0")
+addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.14.6")
 
 addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.6.1")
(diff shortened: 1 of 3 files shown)


project/plugins.sbt
```diff
@@ -4,6 +4,6 @@ addSbtPlugin("com.github.alonsodomin" % "sbt-spark" % "0.6.0")
 
 addSbtPlugin("ch.jodersky" % "sbt-jni" % "1.4.1")
 
-addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.15.0")
+addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.14.6")
```
octaviansima, line 7: 0.15.0 introduced some incompatibilities that were causing `build/sbt assembly` to fail.

wzheng: What were the incompatibilities?

wzheng (approved):
A question!

project/plugins.sbt
octaviansima: Was getting `java.lang.NoSuchMethodError: sbtassembly.AssemblyOption.copy$default$14()Lscala/Enumeration$Value;` in the sbtspark package.

PR #200, opened 2021-04-06
Update Installation Doc

ryanleh:
This PR proposes some tweaks to the installation docs.

1. Installing Spark as specified doesn't create the `/opt/spark/work` directory so the user needs to specify this. Additionally, `sudo chmod -R +wx /opt/spark/work` makes this directory only writeable by root which means you have to run everything as sudo (which I don't believe is intended behavior). 
2. The docs should specify that `private_key.pem` needs to be in the Opaque root directory. I moved the order of a few instructions around to make more logical sense and to resolve this.

### DIFF
--- a/docs/src/install/install.rst
+++ b/docs/src/install/install.rst
@@ -21,7 +21,8 @@ After downloading the Opaque codebase, build and test it as follows.
                    sudo mkdir /opt/spark
                    sudo mv spark-3.1.1*/* /opt/spark
                    rm -rf spark-3.1.1*
-                   sudo chmod -R +wx /opt/spark/work # may not be needed, depends on where Spark is installed
+                   sudo mkdir /opt/spark/work
+                   sudo chmod -R a+wx /opt/spark/work
 
                    # Set Spark environment variables in ~/.bashrc
                    echo "" >> ~/.bashrc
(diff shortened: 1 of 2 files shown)


docs/src/install/install.rst
```diff
@@ -30,27 +31,26 @@ After downloading the Opaque codebase, build and test it as follows.
...
-                   openssl genrsa -out private_key.pem -3 3072
-
-3. Change into the Opaque root directory and edit Opaque's environment variables in ``opaqueenv`` if desired. Export Opaque and OpenEnclave environment variables via
+2. Change into the Opaque root directory and edit Opaque's environment variables in ``opaqueenv`` if desired. Export Opaque and OpenEnclave environment variables via
 
    .. code-block:: bash
                    
                    source opaqueenv
                    source /opt/openenclave/share/openenclave/openenclaverc
-
```
octaviansima, line 45: Please add this line back (follows convention we established)

octaviansima (approved):

docs/src/install/install.rst
ryanleh: Woops my bad - fixed!

docs/src/install/install.rst
```diff
@@ -21,7 +21,8 @@ After downloading the Opaque codebase, build and test it as follows.
                    sudo mkdir /opt/spark
                    sudo mv spark-3.1.1*/* /opt/spark
                    rm -rf spark-3.1.1*
-                   sudo chmod -R +wx /opt/spark/work # may not be needed, depends on where Spark is installed
+                   sudo mkdir /opt/spark/work
+                   sudo chmod -R a+wx /opt/spark/work
```
wzheng, line 25: Should we change the instructions here to support downloading Spark to another location? I usually download it to ~/

octaviansima: When I was looking at install guides, the most common locations I saw were /usr/local and /opt. https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.pdf also mentions that /opt is for add-on applications

wzheng: Ok, then let's keep this!

wzheng (approved):

octaviansima (commented):
One more quick thing: please reflect these changes to the docs in https://github.com/mc2-project/opaque/blob/master/.github/scripts/build.sh

Issue #205, opened 2021-04-07
Missing `sharedKey` when following documentation


ryanleh:
In Step 3 of the `Running the interactive shell` section of `Using Opaque SQL` I get the following error when running `edu.berkeley.cs.rise.opaque.Utils.initSQLContext(spark.sqlContext)`:
```
edu.berkeley.cs.rise.opaque.OpaqueException: Cannot begin attestation without sharedKey.
  at edu.berkeley.cs.rise.opaque.RA$.initRA(RA.scala:47)
  at edu.berkeley.cs.rise.opaque.RA$.attestEnclaves(RA.scala:110)
  at edu.berkeley.cs.rise.opaque.Utils$.initSQLContext(Utils.scala:316)
  ... 49 elided
```

This can be fixed by starting the shell with
```
spark-shell --jars ${OPAQUE_HOME}/target/scala-2.12/opaque_2.12-0.1.jar --conf spark.opaque.testing.enableSharedKey=true
```

instead of what's specified in the docs:
```
spark-shell --jars ${OPAQUE_HOME}/target/scala-2.12/opaque_2.12-0.1.jar
```

But I'm assuming that you don't want to include that configuration flag in this section since it shouldn't be used in production. Perhaps you could specify it's necessary for local testing or something? Or am I doing something wrong?

octaviansima:
I think we need stronger documentation to make it explicitly clear that running anything that isn't the Listener class requires a trusted driver in which this flag is enabled (even Listener currently requires this, however). Basically, nothing but `spark-submit --class edu.berkeley.cs.rise.opaque.rpc.Listener ...` should be used in production in the future. @wzheng what do you think?

wzheng:
Yes, we should specify that we only work in the single party setting, and that the driver program must be put on the client side for production usage, and the testing key should not be used. For anything other than production, the testing key can be enabled. We should document this configuration [here](https://mc2-project.github.io/opaque/usage/usage.html). I wonder if we want to also have a specific section on "Deploying in production"? Though we used to say that Opaque is experimental.

octaviansima:
I like the "deploying in production" idea. I think we should also mention that attestation or log verification is not currently supported in there as well.

octaviansima:
@ryanleh when you have time, can you please work on these amendments to the docs? You've been very thorough with finding gaps in them so far.

PR #208, opened 2021-04-08
Revamp Documents

octaviansima:
Resolves https://github.com/mc2-project/opaque/issues/205

This splits up the usage page in the docs into two: running with a trusted driver, and running with an untrusted one (using MC2 Client). It also creates a default key in `initOpaqueSQL` (previous `initSQLContext`), that would need to be set by loading in a key file from the config later.


docs/src/usage/usage.rst
```diff
@@ -1,21 +1,11 @@
...
-#####
-Opaque SQL needs three Spark properties to be set:
-
-- ``spark.executor.instances``
-- ``spark.task.maxFailures``
--  ``spark.driver.defaultJavaOptions="-Dscala.color"`` (if running the gRPC listener)
-
-These properties can be be set in a custom configuration file, the default being located at ``${SPARK_HOME}/conf/spark-defaults.conf``, or as a ``spark-submit`` or ``spark-shell`` argument: ``--conf <key>=<value>``.
+**************************
+Querying with Spark Driver
```
wzheng, line 2: Suggestion: "Query submission via the Spark driver".

Also: I don't think the driver is defined anywhere? Might be good to include a diagram in the front page, for example from the Opaque paper.

src/main/scala/edu/berkeley/cs/rise/opaque/Utils.scala
```diff
@@ -291,21 +291,25 @@ object Utils extends Logging {
...
+    val sqlContext = spark.sqlContext
     sqlContext.experimental.extraOptimizations =
       (Seq(EncryptLocalRelation, ConvertToOpaqueOperators) ++
         sqlContext.experimental.extraOptimizations)
     sqlContext.experimental.extraStrategies = (Seq(OpaqueOperators) ++
       sqlContext.experimental.extraStrategies)
 
+    // TODO: replace this with loading in a key from a file specified in conf.
+    sharedKey = Some(Array.fill[Byte](GCM_KEY_LENGTH)(1))
+
```
wzheng, line 304: Does this mean the `sharedKey` will always need to be set with a user-specified key in the future? How can users run Opaque in test mode?

docs/src/usage/usage.rst
```diff
@@ -43,60 +33,38 @@ Running the interactive shell
...
-
-2. To launch the listener on a standalone Spark cluster:
-
-   .. code-block:: bash
-
-                  build/sbt assembly # create a fat jar with all necessary dependencies
-
-                  spark-submit --class edu.berkeley.cs.rise.opaque.rpc.Listener  \
-                     <Spark configuration parameters> \
-                     --deploy-mode client ${OPAQUE_HOME}/target/scala-2.12/opaque-assembly-0.1.jar
```
wzheng, line 91: Is this no longer needed? I thought this needs to stay because we also want to interface with the mc2 client?

octaviansima: Forgot to add the new page for this to Git

src/main/scala/edu/berkeley/cs/rise/opaque/Utils.scala
octaviansima: For test mode, there is still the `spark.opaque.testing.enableSharedKey` config.

wzheng: Per our offline conversation: would need to add a "read key from file" functionality here; also need to put that in the documentation.

docs/src/usage/client.rst
```diff
@@ -0,0 +1,41 @@
+********************************************
+Query submission via the MC\ :sup:`2` Client
+********************************************
+
+Starting Opaque SQL
+###################
+
+Here we outline how to use `the client <https://github.com/mc2-project/mc2>`_ to submit queries. In this security model, the Spark driver is considered *untrusted*, and should not have any way to decrypt results or any intermediate computation.
+
+**Note** that this method of computation is still not entirely supported. It is possible to submit queries from MC\ :sup:`2` Client, but only with a shared key on the driver, so it must still be considered *trusted*.
```
wzheng, line 10: This is a bit confusing, since for security reasons the driver still needs to be located on the client side. I would say that this is an alternative way to run Opaque SQL that integrates well with the rest of the mc2 projects, since the mc2-client also works with Secure XGBoost?

src/main/scala/edu/berkeley/cs/rise/opaque/Utils.scala
```diff
@@ -301,11 +303,20 @@ object Utils extends Logging {
     val enableSharedKey = sqlContext.getConf("spark.opaque.testing.enableSharedKey", "false")
```
wzheng, line 303: Shouldn't the default value be true, in case the user forgets to set it?

octaviansima: It depends on what we want the behavior to be. If the code should always default to using a fake key, then yes -- but this means that exceptions such as "cannot encrypt/decrypt without sharedKey" will never occur. I think it makes sense to differentiate between testing case and actual in-practice case, which is why I defaulted to false. The user should always either be explicitly running tests or supplying their own key; anything else would result in error.

wzheng: Oops, yes, you're right -- for I thought enabling shared key means that the user needs to provide it. What about renaming it to `enableTestingKey`, or `enableDummySharedKey`?

src/main/scala/edu/berkeley/cs/rise/opaque/Utils.scala
```diff
@@ -291,21 +292,31 @@ object Utils extends Logging {
...
+    val sqlContext = spark.sqlContext
     sqlContext.experimental.extraOptimizations =
       (Seq(EncryptLocalRelation, ConvertToOpaqueOperators) ++
         sqlContext.experimental.extraOptimizations)
     sqlContext.experimental.extraStrategies = (Seq(OpaqueOperators) ++
       sqlContext.experimental.extraStrategies)
 
-    val enableSharedKey = sqlContext.getConf("spark.opaque.testing.enableSharedKey", "false")
+    val enableSharedKey = sqlContext.getConf("spark.opaque.testing.enableTestingKey", "false")
     if (enableSharedKey.toBoolean) {
```
wzheng, line 304: Nit: should the variables be named `enabledTestingKey` as well?

wzheng (approved):
Just one small comment, otherwise looks good -- thanks!

PR #212, opened 2021-04-19
Add warning for float->string conversions.

### LINKED ISSUE #211: Error deserializing float to string
To demonstrate the error, I can simply follow the instructions from `Using Opaque SQL` in the documentation but substitute the given integers for floats:
```
scala> df.show()
+----+------+
|word| count|
+----+------+
| foo|508.41|
| bar|717.13|
| baz| 82.31|
+----+------+


scala> df.printSchema
root
 |-- word: string (nullable = true)
 |-- count: float (nullable = false)
```
If I encrypt the dataframe, then decrypt it I get the following:
```
scala> dfEncrypted.show()
+----+----------+
|word|     count|
+----+----------+
| foo|508.410004|
| bar|717.130005|
| baz| 82.309998|
+----+----------+


scala> dfEncrypted.printSchema
root
 |-- word: string (nullable = true)
 |-- count: float (nullable = false)


scala> dfEncrypted.collect()
res18: Array[org.apache.spark.sql.Row] = Array([foo,508.41], [bar,717.13], [baz,82.31])
```

So it appears that there is an error in de-serializing the floats to a string as the displayed numbers are incorrect when using `show()` but not `collect()`.

One thing I noticed when debugging is that, if I set breakpoints in the various cases [here](https://github.com/mc2-project/opaque/blob/e0f199c7b28c8550b7c9ac3bde0b8d2318327998/src/main/scala/edu/berkeley/cs/rise/opaque/Utils.scala#L722), running `collect()` shows that both the `StringField` and `FloatField` cases are entered (as expected), but running `show()` prints out that `StringField` is visited twice so it would seem that the float value is being turned into a string (incorrectly) somewhere in C++ code before Scala? But I am not sure.

ryanleh:
Closes #211

### DIFF
--- a/docs/src/usage/functionality.rst
+++ b/docs/src/usage/functionality.rst
@@ -21,6 +21,8 @@ Out of the existing `Spark SQL types <https://spark.apache.org/docs/latest/sql-r
 - ``TimestampTime``, ``DateType``
 - ``ArrayType``, ``MapType``
 
+*Warning:* Conversions from ``FloatType`` to ``StringType`` may produce incorrect results in the least significant portion of the floating point value.  See `this issue <https://github.com/mc2-project/opaque/issues/211>`_ for further details.
+
 Functions
 *********
 


docs/src/usage/functionality.rst
```diff
@@ -21,6 +21,8 @@ Out of the existing `Spark SQL types <https://spark.apache.org/docs/latest/sql-r
 - ``TimestampTime``, ``DateType``
 - ``ArrayType``, ``MapType``
 
+*Warning:* Conversions from ``FloatType`` to ``StringType`` may produce incorrect results in the least significant portion of the floating point value.  See `this issue <https://github.com/mc2-project/opaque/issues/211>`_ for further details.
```
octaviansima, line 24: I don't think "incorrect" is the right word to use here; after reading the issue, it seems that they're both the same float representation. Maybe something like "may produce a correct result rounded to a different decimal place than the one shown by `FloatType`"? But I also kind of don't think this warning is really necessary, since that should be pretty obvious when comparing `show` and `collect` results for example.

ryanleh: Agreed that "incorrect" is the wrong word. Maybe the rephrase could be:
 
 "Converting a ``FloatType`` to ``StringType`` may display a result rounded to a different decimal place than the one shown by ``FloatType`` even though the internal representations are identical."

I don't know if it's good to assume a user would run both `show` and `collect`. Even if they did, I it's very confusing if there was a discrepancy between the two. Thoughts @wzheng ?

wzheng: Another point to consider is that `show` only has "inconsistent behavior" when a user explicitly compares the results to Spark SQL. Opaque's behavior is correct as a standalone SQL engine. So I'd add this point as well to specify what we're comparing to.

PR #215, opened 2021-04-22
Full outer join (via NonObliviousSortMergeJoin)

saharshagrawal:
Added support for FullOuterJoin via NonObliviousSortMergeJoin.

Tested in SGX simulate mode using:
```
build/sbt 'test:testOnly *SinglePartitionJoinSuite -- -t "full outer join"'
build/sbt 'test:testOnly *MultiplePartitionJoinSuite -- -t "full outer join"'
```

Added support for adding a dummy row for the primary table (in addition to having a dummy row for the secondary table) by modifying `strategies.scala` and `operators.scala`. This additional dummy row is needed in the case that a foreign table row appears before any primary table row (implying no matches) and thus we need to append a row containing `primary row nulls + foreign table row`


src/enclave/Enclave/NonObliviousSortMergeJoin.cpp
```diff
@@ -81,10 +80,27 @@ void non_oblivious_sort_merge_join(uint8_t *join_expr, size_t join_expr_length,
...
     const tuix::Row *current = r.next();
+    
     if (current->is_dummy()) {
-      last_foreign_row.set(current);
+      if (join_type == tuix::JoinType_FullOuter) {
+        if (any_primary_row.get() == nullptr) {
+          // Primary table dummy row must come first
+          any_primary_row.set(current);
+        } else if (last_foreign_row.get() == nullptr) {
+          // Foreign table dummy row must come second
```
octaviansima, line 96: I would make it more explicit that this condition is enforced by the Scala code to avoid confusion.

src/enclave/Enclave/NonObliviousSortMergeJoin.cpp
```diff
@@ -81,10 +80,27 @@ void non_oblivious_sort_merge_join(uint8_t *join_expr, size_t join_expr_length,
   // so last_foreign_row.get() is guaranteed to not be null.
   FlatbuffersTemporaryRow last_foreign_row;
 
+  // A "dummy" row needed for FullOuter joins so that 
+  // any_primary_row.get() is guaranteed to not be null
+  FlatbuffersTemporaryRow any_primary_row;
```
octaviansima, line 85: It looks like in any case, whenever we write to output using `last_foreign_row` or `any_primary_row`, their fields are all nulled. Could you update these variables to be something like `dummy_foreign_row` and `dummy_primary_row`? And also remove the call to `last_foreign_row.set(current);` when first encountering a new foreign row. I think it would help to improve readability.

src/main/scala/edu/berkeley/cs/rise/opaque/strategies.scala
```diff
@@ -117,6 +117,12 @@ object OpaqueOperators extends Strategy with JoinSelectionHelper {
             foreignProjSchema.map(_.toAttribute),
             EncryptedSortExec(sortOrder, false, partitioned)
           )
+        case FullOuter =>
+          EncryptedAddDummyRowExec(
+            foreignProjSchema.map(_.toAttribute),
+            EncryptedSortExec(sortOrder, false, partitioned),
+            primaryProjSchema.map(_.toAttribute)
```
octaviansima, line 124: Is it possible to achieve this additional functionality without modifying `EncryptedAddDummyRowExec`? I would try having two calls to `EncryptedAddDummyRowExec` in this `FullOuter` case.

src/test/scala/edu/berkeley/cs/rise/opaque/JoinSuite.scala
```diff
@@ -172,7 +172,7 @@ trait JoinSuite extends OpaqueSQLSuiteBase with SQLHelper {
     }
   }
 
-  ignore("full outer join") {
+  test("full outer join") {
```
octaviansima, line 175: Can you try adding the full outer join queries in `unsupportedQueries` to `queries` as well please?

octaviansima (commented):
Thanks for the PR! Waiting for @wzheng to take a look as well, but I added some comments.

src/main/scala/edu/berkeley/cs/rise/opaque/strategies.scala
saharshagrawal: Yeah, I wasn't sure about doing it this way either, so I'll try the other approach instead.

src/enclave/Enclave/NonObliviousSortMergeJoin.cpp
saharshagrawal: Sure, will do

src/enclave/Enclave/NonObliviousSortMergeJoin.cpp
saharshagrawal: Ok, I'll add better comments and/or try to refactor

src/test/scala/edu/berkeley/cs/rise/opaque/JoinSuite.scala
saharshagrawal: I think some of those are non-equi joins, so I was going to wait until finishing the BNLJ implementation before adding those.

src/enclave/Enclave/NonObliviousSortMergeJoin.cpp
```diff
@@ -137,6 +153,9 @@ void non_oblivious_sort_merge_join(uint8_t *join_expr, size_t join_expr_length,
               } else {
                 w.append(primary, current);
               }
+            } else if (join_type == tuix::JoinType_FullOuter){
+              // Join condition not satisfied; need to add nulls ++ foreign row to output
+              w.append(any_primary_row.get(), current, true, false);
```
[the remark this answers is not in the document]
saharshagrawal: Yeah, you're right; I'll add a flag to detect if any matches occurred and move the logic to outside the while loop and only add a (nulls, foreign row) if the flag is false.

saharshagrawal:
Build should succeed now; previously there were compilation bugs that I didn't detect due to local configuration issues

src/enclave/Enclave/NonObliviousSortMergeJoin.cpp
```diff
@@ -120,25 +137,31 @@ void non_oblivious_sort_merge_join(uint8_t *join_expr, size_t join_expr_length,
...
       }
     } else {
-      last_foreign_row.set(current);
       if (last_primary_of_group.get()
           && join_expr_eval.is_same_group(last_primary_of_group.get(), current)) {
         if (join_type == tuix::JoinType_Inner || join_expr_eval.is_outer_join()) {
           auto primary_group_buffer = primary_group.output_buffer();
           RowReader primary_group_reader(primary_group_buffer.view());
 
+          bool matchFound = false;
```
octaviansima, line 146: Use snake case for consistency

octaviansima (approved):
LGTM! Thanks for implementing these changes. Please run `./format.sh` to lint changed files; we'll merge it in after that.

src/enclave/Enclave/NonObliviousSortMergeJoin.cpp
saharshagrawal: fixed

saharshagrawal:
> LGTM! Thanks for implementing these changes. Please run `./format.sh` to lint changed files; we'll merge it in after that.

Should be done now; thought I had installed the pre-commit hook, but must've accidentally been removed.

PR #221, opened 2021-05-05
Reorganize C++ code

octaviansima:
This PR does 4 main things:

1. Rename all CamelCase C++ files to snake_case, which is the usual convention for C/C++ files.
2. Clean up the CMake files and have them use modern CMake operations such as `target_compile_options` instead of explicitly setting CMAKE_CXX_FLAGS in the root CMake. This will have the benefit of easier debugging down the road, since it reduces the scope of common flags to those targets that absolutely need them.
3. Reorganize the `cpp/enclave` files into three main directories: crypto/, flatbuffer_helpers/, and physical_operators/. Previously, all files were in the same directory, regardless of purpose. This change increases readability and defines a more logical trusted directory structure.
4.   Delete unused files: ServiceProvider/KeyGen.cpp, enclave/benchmark.sh, enclave/build.sh, enclave/fetch-data.sh


src/cpp/enclave/CMakeLists.txt
```diff
@@ -0,0 +1,74 @@
+cmake_minimum_required(VERSION 3.13)
+
+project(OpaqueEnclaveTrusted)
+
+set(SOURCES
+  crypto/ks_crypto.cpp
+  crypto/m_crypto.cpp
```
wzheng, line 7: It's a bit unclear how these two crypto files are divided up -- why are there two crypto files? Maybe we should merge them into one?

ryanleh: It might not be worth it to spend effort merging crypto code rn since I'm planning to create the Crypto/Attestation shared repo in the near future.

octaviansima: m_crypto.cpp is a modified file from OE as well -- I think we should keep them separate for easier version control as well in case we need to update it.

PR #223, opened 2021-05-17
Add issue templates

ryanleh:
This PR adds issue templates for bugs and feature requests.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,26 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+<!-- < < < < < < < < < < < < < < < < < < < < < < < < < < < < < < < < < ☺ 
+v                            ✰  Thanks for opening an issue! ✰    
+v    Before smashing the submit button please review the template.
+v    Please also ensure that this is not a duplicate issue :)  
+☺ > > > > > > > > > > > > > > > > > > > > > > > > > > > > > > > > >  -->∂
+
+## Summary of Bug
+
+<!-- Concisely describe the issue -->
+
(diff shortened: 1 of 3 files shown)


chester-leung:
Should we encourage people to use issue templates by[ not enabling blank issues by default](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#configuring-the-template-chooser)?

"You can customize the issue template chooser that people see when creating a new issue in your repository by adding a config.yml file to the .github/ISSUE_TEMPLATE folder.

You can encourage contributors to use issue templates by setting blank_issues_enabled to false. If you set blank_issues_enabled to true, people will have the option to open a blank issue."

wzheng:
If we disable blank issues, then we'd have to enable more templates than these. For example, some people create Github issues for setup problems, which are not bugs or feature requests. I think we should have at least another "Support" template. What do you think, @octaviansima?

octaviansima:
I think we should keep blank issues for the case when an issue does not match a certain template, rather than create templates for all those various scenarios.

ryanleh:
Ok, so is there anything left to do on this then?

wzheng:
I agree with Octavian on enabling blank issues, though I think it would be nice to add a "Support" template?

ryanleh:
Ok, I added a support template. Let me know if this is what you meant.

chester-leung:
Will the support template now be the default template when opening an issue? Or how will that work?

ryanleh:
So I'm not sure how to test this but, going off of this example [here](https://github.com/arkworks-rs/poly-commit), I believe users will choose one of them when they open an issue.

chester-leung:
Ok this looks good then!

chester-leung (approved):

ryanleh:
Ok, everything looks like it's working as intended! 🎉

PR #224, opened 2021-05-17
Add support for oegdb

octaviansima:
This PR fixes the previously broken SBT task to run a test suite with a debugger. A couple of things to note:
1. Changing the test suite to run requires modifying project/resources/run-tests.gdb. I tried to make it so it can be supplied through the commandline like `build/sbt 'oe-gdb edu.berkeley.cs.rise.opaque.SinglePartitionJoinSuite'`, but the actual SBT task requires a modified state created in `oeGdbCommand`. It is only possible to return a modified state from a task (even this is deprecated), not modify it mid task. In order to parse in command-line arguments (`edu.berkeley.cs.rise.opaque.SinglePartitionJoinSuite` in this case), we need to use `InputKey` instead of `TaskKey`. The problem is that `runInputTask` accepts a different set of inputs than the ones specified in the commandline, and it is not possible to access commandline arguments through the parser outside of an input task definition (in this case `oeGdbCommand`).
2. All javaOptions for sbt-launch.jar in build.sbt are currently not being set. The custom sbt runner we use, build/sbt, defines its own set of Java options in `default_jvm_opts_common`. This runner allows us to specify a file for custom JVM options, .jvmopts, but we cannot set command/task-specific JVM options through this (`javaOptions in oe-gdb`, which we need). For GDB to work, we need to set `-Dsbt.supershell=never` because otherwise the commands to gdb through `build/sbt oe-gdb` continuously get overridden by a terminal refresh, making reading what was already written impossible.
3. There is an assertion in `ocall_malloc` that is failing. I disabled this only for the DEBUG case, but it is strange because `oe_is_outside_enclave` is passing in all other cases.

Planning on making tasks for all these to fix later, but for now GDB works and I think we have other priorities. Let me know if you have concerns though.

### DIFF
--- a/.jvmopts
+++ b/.jvmopts
@@ -1,0 +1,1 @@
+-Dsbt.supershell=never
--- a/build.sbt
+++ b/build.sbt
@@ -162,12 +162,12 @@ watchSources ++=
 
 val synthTestDataTask = TaskKey[Unit]("synthTestData", "Synthesizes test data.")
 
-val sgxGdbTask =
-  TaskKey[Unit]("sgx-gdb-task", "Runs OpaqueSinglePartitionSuite under the sgx-gdb debugger.")
-def sgxGdbCommand = Command.command("sgx-gdb") { state =>
+val oeGdbTask =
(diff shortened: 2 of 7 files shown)


wzheng (approved):
Thanks, this is great!!

PR #228, opened 2021-05-26
Add install script

octaviansima:
This PR adds an install script (`scripts/install.sh`) to automatically install all dependencies and generate necessary files (keys). It also contains changes to the documentation to show how to use this script instead of manually installing everything.


scripts/install.sh
```diff
@@ -0,0 +1,65 @@
...
+sudo mkdir /opt/spark/work
+sudo chmod -R a+wx /opt/spark/work
+
+# Set Spark environment variables in ~/.bashrc
+echo "" >> ~/.bashrc
+echo "# Spark settings" >> ~/.bashrc
+echo "export SPARK_HOME=/opt/spark" >> ~/.bashrc
+echo "export PATH=$PATH:/opt/spark/bin:/opt/spark/sbin" >> ~/.bashrc
+echo "" >> ~/.bashrc
+source ~/.bashrc
```
wzheng, line 55: Should this be part of bashrc or opaqueenv?

wzheng (approved):

PR #229, opened 2021-06-02
Update LICENSE

octaviansima:
This updates the LICENSE file to include files from the Linux SGX extensions. Also updates the headers to use the new (open-sourced) license.


wzheng:
Is there any OpenEnclave that is used?

octaviansima:
> Is there any OpenEnclave that is used?

There was in m_crypto.cpp, but that is being deleted after https://github.com/mc2-project/opaque-sql/pull/233 is merged.

wzheng (approved):

PR #231, opened 2021-06-02
Add static signing key

ryanleh:
This PR adds a static signing key and symmetric key to Opaque SQL. The public part of the signing key will be added to the `mc2-client` so that attestation will work out-of-the-box for demos. The docs have been also been updated to warn against using these keys in production.


docs/src/install/install.rst
```diff
@@ -44,7 +44,7 @@ After downloading the Opaque codebase, build and test it as follows.
    By default, Opaque runs in hardware mode (environment variable ``MODE=HARDWARE``).
    If you do not have a machine with a hardware enclave but still wish to test out Opaque's functionality locally, then set ``export MODE=SIMULATE``.
 
-3. There is an ``sbt`` task you can use to generate private and symmetric keys using OpenSSL for Opaque SQL.
+3. By default, Opaque SQL uses a pre-generated RSA private key and symmetric key. These keys _should be not be used in a production environment_. To generate new keys using OpenSSL, you can use the following ``sbt`` task:
```
octaviansima, line 47: There is an empty space after environment. Please also specify that with other keys, the `PRIVATE_KEY_PATH` and `SYMMETRIC_KEY_PATH` environment variables set in `opaqueenv` will have to be reset as well.

octaviansima:
Are these keys just to communicate securely with the client?

docs/src/install/install.rst
ryanleh: Thanks for the catch - fixed

ryanleh:
Ya, this is essentially to make sure that Opaque SQL works with the client out of the box without needing to move keys between the two repos. 

Alternatively, we could disable attestation by default but that doesn't seem like the greatest. Or perhaps there is another solution?

octaviansima:
Hmm, I think that `PRIVATE_KEY_PATH` is required for that, but `SYMMETRIC_KEY_PATH` shouldn't be: this is the key that a *trusted* driver uses to decrypt/re-encrypt intermediate results, and will not exist when interfacing with the client. Attestation will be disabled by default during tests after https://github.com/mc2-project/opaque-sql/pull/233, so maybe a cleaner way to do it is to transfer the certificate during attestation?

ryanleh:
Ya so I guess I included it so that a user wouldn't need to run `build/sbt keys` by default and security didn't really matter since running it without a freshly-generated signing key would be insecure anyways. But I can change that back.

Gotcha. I don't see how transferring a certificate would help since you still need to know the public key of the certificate signer (or maybe I'm misunderstanding). 

What do we think... (adding @chester-leung @wzheng) should we have this static signing key by default, or just disable attestation in tests and demos by default? I might lean towards the latter honestly and just make sure that we display a very verbose error so people remember to enable it for an actual production use.

docs/src/install/install.rst
```diff
@@ -44,13 +44,15 @@ After downloading the Opaque codebase, build and test it as follows.
...
+3. To generate new keys using OpenSSL, you can use the following ``sbt`` task:
 
    .. code-block:: bash
 
                   build/sbt keys
 
-   Alternatively, you can use your own keys, though this is generally not recommended.
+   *Note that, by default, Opaque SQL uses a pre-generated RSA private key for enclave signing. This key should be not be used in a production environment.* 
+
+   Alternatively, you can use your own keys, though this is generally not recommended.
```
octaviansima, line 55: The key that `PRIVATE_KEY_PATH` points to has to be identical to the one on the client, correct? If that's the case then I don't think we need this SBT task anymore.

ryanleh: The client won't have the private key, but the client should have the public key corresponding to the key at `PRIVATE_KEY_PATH`. Why wouldn't we need the SBT task anymore? Doesn't it still generate the symmetric key?

octaviansima: lol too many keys

Yeah it does, but that symmetric key will be useless after https://github.com/mc2-project/opaque-sql/pull/233 is merged in

octaviansima (approved):
Thanks!

docs/src/install/install.rst
ryanleh: Sounds good 😄

PR #235, opened 2021-06-07
Update Dockerfile + add Docker build instructions

chester-leung:
This PR updates the Dockerfile to build and install the most recent version of Opaque SQL, and adds instructions to the README on how to build the Docker image and launch a container from it.


chester-leung:
Never mind, combining this with the client Dockerfile.

chester-leung:
Reopening, in case we want to replace the image in the current quickstart with a much faster to build Dockerfile

PR #242, opened 2021-06-17
Change accent color

ryanleh:
This changes the accent color of the Opaque SQL docs to match the client -- the current color is a bit harsh.

Before:
Screen Shot 2021-06-17 at 12 04 09 PM

After:
Screen Shot 2021-06-17 at 12 09 01 PM


octaviansima (approved):
LGTM!

PR #248, opened 2021-07-07
Add Pyspark listener

octaviansima:
This PR adds support for a Pyspark listener that is intended to receive queries from the [MC<sup>2</sup> Client](https://github.com/mc2-project/mc2). It contains SBT code to generate Python protobuf files and package them into a zip automatically for submitting to a standalone Spark cluster. The instructions for running the listener are in the client docs.

Unfortunately, it also introduces a Python dependency to Opaque SQL, since there are required packages to start the Python listener service.

S242


src/python/intp_handler.py
```diff
@@ -0,0 +1,21 @@
+import code
+import io
+from code import InteractiveInterpreter
+from contextlib import redirect_stdout, redirect_stderr
+
+from pyspark import SparkConf, SparkContext
+
+class IntpHandler:
```
wzheng, line 8: I'm looking at the pyspark shell code (https://github.com/apache/spark/blob/master/python/pyspark/shell.py) -- does it work if you just call `exec(code_string)` directly, instead of using an `InteractiveInterpreter`? Since it's Python, calling `exec` should just run the user's code.

octaviansima: Good point, that should work. I was unsure if `exec` uses the state from the main Python program but it looks like that is the case: https://www.programiz.com/python-programming/methods/built-in/exec

wzheng (approved):

PR #263, opened 2022-02-18
Fix `OPAQUE_HOME` environment variable

octaviansima:
Using PWD doesn't work if sourced outside of the root repo.


chester-leung (approved):

chester-leung:
LGTM! Thanks.

